PR #67, opened 2021-06-09
Local Development Environment For The Gateway

derrandz:


.github/workflows/staging-us-west-2.yml
```diff
@@ -2,7 +2,7 @@ name: Staging Deployment us-west-2
 
 on:
   push:
-    branches: [ master-off ]
+    branches: [ master ]
```
nymd: this should remain off

nymd (changes requested):

.github/workflows/staging-us-west-2.yml
derrandz: but we want to deploy to staging right?

.github/workflows/testnet.yml
```diff
@@ -2,7 +2,7 @@ name: Testnet Deployment
 
 on:
   push:
-    branches: [ staging-off ]
+    branches: [ staging ]
```
nymd, line 5: this should be branches: [ testnet ] 
or something. There is a real testnet deployment that uses pocket testnet and replicates the production setup.

I don't want this deployment to receive staging code unless we purposefully want it there.

nymd (changes requested):

nymd (approved):

PR #100, opened 2021-07-02
New altruist implementation

nymd:
- Replaced "FALLBACK_URL" env variable with "ALTRUISTS"; this is reflected in the production configs and local .env.example
- Created [Altruist infra](https://lucid.app/publicSegments/view/2294b8cb-0741-404a-b59e-91dc2fe78b9e/image.png) with HTTPS load balancer, nginx, and multiple blockchain nodes of each type
- Fixed a few minor bugs introduced from development environment
- Removed local blockchain-subdomain as localhost can work fine with subdomains
- Upgraded Loopback 4 to latest; this introduced a possible need for a new sequence.ts #102  and required the manual addition of CORS headers to make it work for now


src/application.ts
```diff
@@ -47,9 +46,11 @@ export class PocketGatewayApplication extends BootMixin(
     // Requirements; for Production these are stored in GitHub repo secrets
     //
     // For Dev, you need to pass them in via .env file
-    const environment: string = process.env.NODE_ENV || 'production';
+    let environment: string = process.env.NODE_ENV ?? 'production';
```
Evalir: Any reason we prefer `let` to `const`?

src/services/pocket-relayer.ts
```diff
@@ -9,37 +9,19 @@ import {
...
-  headers: null;
-}
-interface FallbackMeta {
-  block_height: number;
-}
-interface FallbackProof {
-  blockchain: String;
-}
+const logger = require('../services/logger');
+const axios = require('axios');
```
Evalir: Can't we use `import` syntax here?

src/services/pocket-relayer.ts
```diff
@@ -235,82 +206,86 @@ export class PocketRelayer {
...
-      };
-
-      const fallbackResponse = await fallbackChoice.send(
-        '/v1/client/relay',
-        JSON.stringify(fallbackRelay),
-        1200000,
-        false,
-      );
+      // Add relay path to URL
+      const altruistURL = (relayPath === undefined || relayPath === "") ? this.altruists[blockchain] : this.altruists[blockchain] + '/' + relayPath;
```
Evalir: We could probably make this use more modern syntax. Roughly:
```suggestion
      const altruistURL = (relayPath === undefined || relayPath === "") ? this.altruists[blockchain] : `${this.altruists[blockchain]}/${relayPath};
```

src/services/pocket-relayer.ts
```diff
@@ -235,82 +206,86 @@ export class PocketRelayer {
...
-        const responseParsed = JSON.parse(fallbackResponse);
+      try {
+        const fallbackResponse = await axios(axiosConfig);
+
+        if (this.checkDebug) {
+          logger.log('debug', JSON.stringify(fallbackResponse.data), {
+            requestID: requestID,
+            relayType: 'FALLBACK',
+            typeID: application.id,
+            serviceNode: 'fallback:' + redactedAltruistURL,
```
Evalir, line 242: Same thing here and in the other places below:
```suggestion
            serviceNode: `fallback:${redactedAltruistURL}`,
```

Evalir (approved):
Looks good! Feel free to ignore if this must land quickly, but just some syntax suggestions:

src/application.ts
nymd: Oops. Vestigial `let` from when I was trying to diagnose why it wasn't accepting my environment variable as being production.

src/services/pocket-relayer.ts
nymd: I'm old and my concatenation style is old.

src/services/pocket-relayer.ts
nymd: tbh I do this everywhere in the code. It will have to be prettified later. Hopefully automatically.

derrandz:
Blockchain subdomains were not working in my machine. At that time I was on Ubuntu and the portal was running from inside of a container.

@nymd was it as simple as http://blockchain-domain.localhost:port ?

Issue #124, opened 2021-07-08
Prettify code


nymd:
The current code could be automatically prettified. 

There are also a number of functions with very long lists of variables that are passed. Those could be changed into destructured objects to make things easier to pass through.

There's lots of cosmetic improvements that could be made to the code.

nymd:
Would be great to have the prettifier/linter run in a hook every time you try and push

PR #136, opened 2021-07-08
Modify prettier configuration and run on the codebase

rem1niscence:
The prettier configuration, although set up on the project, never actually ran as was supposed to be intended, this was due to two reasons:

* The file is incorrectly place on the project (inside the `pocket-gateway` folder) when the [configuration docs](https://prettier.io/docs/en/configuration.html) states that the checked files **look up** a prettier configuration file, as this was isolated on the `pocket-gateway` folder, any file wouldn't find it and instead a default configuration of prettier is being run.
* The command `npm run prettier:fix` is internally running a loopback configuration of prettier, which ignores the project's `prettier` and `.prettierignore` so not even placing the prettier config at root would solve the problem.

The fix is to move the prettier files to root and change the npm script to run the native's prettier command, after that the command is ran and applies all the code style changes. Leading to this **HUGE** PR which practically touches all the files, think of this as the first time prettier was actually ran.

This kinda touches on the issue #124 but only on the prettify code, doesn't refactor any functionality by destructuring or any other technique, I didn't do that as this PR is already way too big as it is right now. Will do it on a separate PR.


.prettierrc
```diff
@@ -0,0 +1,8 @@
+{
+  "trailingComma": "es5",
+  "tabWidth": 2,
+  "semi": false,
+  "singleQuote": true,
+  "bracketSpacing": true,
+  "arrowParens": "always"
```
nymd: 120 line length pls

nymd (changes requested):

.prettierrc
rem1niscence: Done

nymd (approved):

PR #155, opened 2021-07-09
Fix eslint configuration and precommit hooks to lint checks

rem1niscence:
Just like #136 the eslint configuration wasn't properly set up and wasn't running as expected, the same fix as prettier was applied  and fixed all the warnings errors by also adding a more exhaustive config. A precommit hook was added using [husky](https://github.com/typicode/husky) which for now only checks for prettier and eslint but can be extended


.eslintrc.js
```diff
@@ -0,0 +1,76 @@
+module.exports = {
+  root: true,
+  env: {
+    es6: true,
+    mocha: true,
```
Evalir: Just FYI—if we're considering using Jest for tests then we might need to change this later on.

package.json
```diff
@@ -99,8 +101,21 @@
...
     "tsnodemon": "^1.2.2",
     "typescript": "~4.2.4"
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged"
+    }
+  },
+  "lint-staged": {
+    "*.(t|j)s": [
```
Evalir, line 116: Sweet regex!

src/application.ts
```diff
@@ -127,8 +130,8 @@ export class PocketGatewayApplication extends BootMixin(ServiceMixin(RepositoryM
     }
 
     // Load Redis for cache
-    const redisEndpoint: string = process.env.REDIS_ENDPOINT || ''
-    const redisPort: string = process.env.REDIS_PORT || ''
+    const redisEndpoint: string = process.env.REDIS_ENDPOINT ?? ''
+    const redisPort: string = process.env.REDIS_PORT ?? ''
```
Evalir: Actually thinking that these `??` might not be needed as we're checking if the value's falsy below. This was `ESLint` complaining though right?

src/services/pocket-relayer.ts
```diff
@@ -309,22 +313,37 @@ export class PocketRelayer {
...
+    requestTimeOut,
+    blockchain,
+    blockchainEnforceResult,
+    blockchainSyncCheck,
+    blockchainSyncCheckPath,
+    blockchainSyncAllowance,
+    blockchainSyncBackup,
+    blockchainIDCheck,
+    blockchainID,
+  }: {
```
Evalir, line 331: Note that adding an object changes the function signature. Were all calls to this function also updated?

src/utils/enforcement.ts
```diff
@@ -4,7 +4,9 @@ export function checkEnforcementJSON(test: string): boolean {
   }
   // Code from: https://github.com/prototypejs/prototype/blob/560bb59414fc9343ce85429b91b1e1b82fdc6812/src/prototype/lang/string.js#L699
   // Prototype lib
-  if (/^\s*$/.test(test)) return false
+  if (/^\s*$/.test(test)) {
+return false
+}
```
Evalir: This probably needs some indenting :)

src/services/sync-checker.ts
```diff
@@ -16,41 +17,39 @@ export class SyncChecker {
...
+    blockchainSyncBackup,
+    applicationID,
+    applicationPublicKey,
+    pocket,
+    pocketAAT,
+    pocketConfiguration,
+  }: ConsensusFilterOptions): Promise<Node[]> {
     // Blockchain records passed in with 0 sync allowance are missing the 'syncAllowance' field in MongoDB
     if (syncAllowance <= 0) {
       syncAllowance = 5
```
Evalir, line 36: not related to the PR, but this is an arbitrary value, right? We should probably put it as a constant at the top of the file (`DEFAULT_SYNC_ALLOWANCE`) even though it's not that obvious what this does at first glance.

src/services/sync-checker.ts
```diff
@@ -16,41 +17,39 @@ export class SyncChecker {
...
+    syncCheckPath,
+    syncAllowance = 5,
+    blockchain,
+    blockchainSyncBackup,
+    applicationID,
+    applicationPublicKey,
+    pocket,
+    pocketAAT,
+    pocketConfiguration,
+  }: ConsensusFilterOptions): Promise<Node[]> {
```
Evalir, line 33: Same as `sendRelay`—are we aware we're changing the whole function signature? have we updated all calls to this function?

src/services/cherry-picker.ts
```diff
@@ -169,11 +176,13 @@ export class CherryPicker {
     const serviceLog = await this.fetchRawServiceLog(blockchain, id)
 
     let serviceQuality
+
     // Update service quality log for this time period
     if (serviceLog) {
       serviceQuality = JSON.parse(serviceLog)
 
       let totalResults = 0
```
Evalir, line 184: Let's remember at some point to turn these into consts :see_no_evil:

src/services/pocket-relayer.ts
```diff
@@ -646,3 +674,14 @@ interface BlockchainDetails {
   blockchainIDCheck: string
   blockchainID: string
 }
+
+export interface SendRelayOptions {
```
Evalir, line 678: Any reason to prefer interface over types? This is fine by me as it's a matter of preference I believe, just wanna make sure we stick to a standard.

Evalir (commented):
In general I'd prefer Alex to sign off on this—I'm just glossing over and nitpicking on the things I see :). My main concern is the function signature changes:

src/services/chain-checker.ts
```diff
@@ -201,24 +209,24 @@ export class ChainChecker {
     ] = await Promise.all(promiseStack)
 
     for (const rawNodeChainLog of rawNodeChainLogs) {
-      if (typeof rawNodeChainLog === 'object' && rawNodeChainLog.chainID !== '') {
+      if (typeof rawNodeChainLog === 'object' && ((rawNodeChainLog.chainID as unknown) as string) !== '') {
```
nymd: Nice

src/services/metrics-recorder.ts
```diff
@@ -33,7 +33,7 @@ export class MetricsRecorder {
   async recordMetric({
     requestID,
     applicationID,
-    appPubKey,
```
nymd, line 36: This was to match the metrics DB but I'm fine either way

src/services/pocket-relayer.ts
nymd: Only called once in the codebase though with retries turned on it could be called multiple times (from the same place)

src/services/sync-checker.ts
nymd: Also only called from one place in the codebase.

src/services/sync-checker.ts
nymd: Agreed, this should be abstracted to a CONST that we set in the environment files (workflows/...yml)

nymd (changes requested):

src/application.ts
rem1niscence: yep, I'll be disabling that rule so is not asked all the time but up to developer's discretion

src/services/pocket-relayer.ts
rem1niscence: and yes, that one place where is called was updated, if not ts would have complained

src/utils/enforcement.ts
rem1niscence: yeah it does, should have been updated with prettier precommit, will look into that

src/services/sync-checker.ts
rem1niscence: can do it here tho, is a simple change, although it wont go with the nature of this PR, let me know if I should not

src/services/pocket-relayer.ts
rem1niscence: yeah, that's definitely a type, good catch! the standard here should be that unless it can't be done by a type (for example inheritance) then use a type

src/services/sync-checker.ts
nymd: That's true, we can wait on that

nymd: https://github.com/pokt-foundation/portal-api/issues/156

src/services/chain-checker.ts
rem1niscence: so, change it or not? 😂😂

nymd: Yes it's fine, just saying I learned something

src/services/metrics-recorder.ts
rem1niscence: yeah I removed unused code, if that's needed then it can be added again but for now 💣

src/services/cherry-picker.ts
rem1niscence: but why? I mean, that variable is changed depending on of the code below so is gotta be a let

nymd (approved):

PR #158, opened 2021-07-13
Change syncAllowance default value as an environment variable

### LINKED ISSUE #156: Change hardcoding of default sync allowance
Currently syncAllowance is defaulted to a hardcoded '5'

This should be changed to a constant like DEFAULT_SYNC_ALLOWANCE and moved to the variables for deployment.

rem1niscence:
* Change sync allowance to a environment variable and change the workflows to reflect the default value
* Change the null coalescing operator as is not supported on the project's node engine

Fixes #156


src/services/sync-checker.ts
```diff
@@ -8,6 +8,8 @@ const logger = require('../services/logger')
 
 import axios from 'axios'
 
+const DEFAULT_SYNC_ALLOWANCE: number = parseInt(process.env.DEFAULT_SYNC_ALLOWANCE) || 5
```
nymd: I'd rather have all of the .env variables checked in one place and rather than silently fallback to a default, I'd rather it error out and tell you that you made a dumb mistake and forgot the var.

Probably around here: https://github.com/pokt-foundation/portal-api/blob/bddc518b3d833a088e787f1823587313b7ba33e1/src/application.ts#L59

nymd (changes requested):

src/services/sync-checker.ts
Evalir: I like having `.env` vars in one place—[this](https://github.com/pokt-foundation/portal/blob/mvp/packages/backend/src/environment.ts) is how I usually do it

nymd: That does look nicer.

rem1niscence: yep, like it too

rem1niscence: for now I'm doing a quick fix but an issue can be opened for that to change all the envs

src/application.ts
```diff
@@ -44,19 +44,20 @@ export class PocketGatewayApplication extends BootMixin(ServiceMixin(RepositoryM
...
-    const aatPlan = process.env.AAT_PLAN ?? AatPlans.PREMIUM
+    const dispatchURL: string = process.env.DISPATCH_URL || ''
+    const altruists: string = process.env.ALTRUISTS || ''
+    const clientPrivateKey: string = process.env.GATEWAY_CLIENT_PRIVATE_KEY || ''
+    const clientPassphrase: string = process.env.GATEWAY_CLIENT_PASSPHRASE || ''
+    const pocketSessionBlockFrequency: string = process.env.POCKET_SESSION_BLOCK_FREQUENCY || ''
+    const pocketBlockTime: string = process.env.POCKET_BLOCK_TIME || ''
+    const relayRetries: string = process.env.POCKET_RELAY_RETRIES || ''
+    const databaseEncryptionKey: string = process.env.DATABASE_ENCRYPTION_KEY || ''
+    const defaultSyncAllowance: number = parseInt(process.env.DEFAULT_SYNC_ALLOWANCE) || -1
```
rem1niscence, line 59: is 0 a valid value or must always be over 0? @nymd

nymd: 0 is valid though it would be very restrictive. It should be allowed. But not blank or undefined.

rem1niscence: all right, then this is ready

nymd (approved):

PR #174, opened 2021-07-21
feat: set blocks range limit on eth_getLogs calls

### LINKED ISSUE #62: eth_getLogs should be restricted by block distance
Currently calls to the Portal API are unmoderated in regards to how many blocks you can call with eth_getLogs. This can cause excess load on relay chain nodes and often results in timeouts.

App developers simply don't realize that you can't call for logs from a busy contract from block 0 to latest in one call.

Competitors that index the chains in a relational database limit the calls by number of rows -- for instance you can't call eth_getLogs for queries that will return more than 10,000 rows.

This isn't possible for us as we don't know beforehand the amount of rows that will return.  We should instead implement a block-based limit that prevents eth_getLogs calls from requesting a range of blocks that is bigger than a given set, say 10,000 blocks.

This is a sample call that is requesting the entire history of a contract:

` ~ $ curl -X POST http://localhost:8545 -H 'User-Agent: pocket-debug' -H 'Content-Type: application/json' -d '{"method":"eth_getLogs","params":[{"fromBlock":"0x0","toBlock":"latest","address":"0x36f73b436db8af70ab6cffc4bbc99c95990168f0"}],"id":256145,"jsonrpc":"2.0"}'`

Relay inspection should be done on the passed call. Analyze the amount of blocks between fromBlock and toBlock. Deny the relay with a meaningful error if the spread is > X blocks. 

- X should be set as a variable in the deployment sections of .github/workflows
- A new service should be created that allows for deep inspection of relays as this situation may come up with other calls/chains

crisog:
Closes #62. Builds on #199.

- [x] Error when request logs for X > 10k blocks.
- [x] Sucess when request logs for X <= 10k blocks.
- [x] Fetches latest block when `toBlock` or `fromBlock` are "latest".
- [x] Uses latest block when `toBlock` is undefined.
- [x] Uses latest block when `fromBlock` is undefined.
- [x] If no altruists are available, an error is returned with a message indicating to use an explicit block number instead of latest.


src/services/pocket-relayer.ts
```diff
@@ -623,6 +635,31 @@ export class PocketRelayer {
...
+      let [{ fromBlock, toBlock }] = parsedRawData.params as [{ fromBlock: string; toBlock: string }]
+
+      // If fromBlock or toBlock is equal to undefined it defaults to latest.
+      if ((fromBlock === 'latest' || fromBlock === undefined) && altruistUrl !== 'undefined') {
+        fromBlock = String(await getBlockNumber(altruistUrl))
+      } else if ((toBlock === 'latest' || toBlock === undefined) && altruistUrl !== 'undefined') {
+        toBlock = String(await getBlockNumber(altruistUrl))
+        // We cannot move forward if there is no altruist available.
+      } else if (
+        (toBlock === 'latest' || toBlock === undefined || fromBlock === 'latest' || fromBlock === undefined) &&
```
rem1niscence: this is hard to read, you could separate these into variables like `toBlockValid` checking them separately so at the end end on the else if you only do a simpler check like `!toBlockValid && !fromBlockValid`

crisog:
Waiting for pocket relayer mock to include tests, @rem1niscence is working on it.

src/application.ts
```diff
@@ -78,6 +79,7 @@ export class PocketGatewayApplication extends BootMixin(ServiceMixin(RepositoryM
     const relayRetries: string = POCKET_RELAY_RETRIES || ''
     const databaseEncryptionKey: string = DATABASE_ENCRYPTION_KEY || ''
     const defaultSyncAllowance: number = parseInt(DEFAULT_SYNC_ALLOWANCE) || -1
+    const defaultLogLimitBlocks: number = parseInt(LOG_LIMIT_BLOCKS) || -1
```
nymd: if the const is 'defaultLogLimitBlocks' then please make the ENV:

DEFAULT_LOG_LIMIT_BLOCKS

src/services/pocket-relayer.ts
```diff
@@ -637,12 +656,80 @@ export class PocketRelayer {
...
+      if (toBlockParam !== undefined && toBlockParam !== 'latest') {
+        toBlock = blockHexToDecimal(toBlockParam)
+        isToBlockHex = true
+      }
+      if (fromBlockParam !== undefined && fromBlockParam !== 'latest') {
+        fromBlock = blockHexToDecimal(fromBlockParam)
+        isFromBlockHex = true
+      }
+
+      if (altruistUrl !== 'undefined') {
```
nymd: Can we only make this check if necessary?  Only if 'latest' is used as a variable. It seems we don't need this call if we know the block numbers already.

src/utils/block.ts
```diff
@@ -0,0 +1,3 @@
+export function blockHexToDecimal(hex: string): number {
```
nymd, line 1: I think there are some instances of this code in sync-checker that could be removed and instead use this central util?

nymd (changes requested):
Looks good, just a few minor things.

src/errors/limit-error.ts
```diff
@@ -0,0 +1,13 @@
+/**
```
rem1niscence: this file in general could have its name changed to `types.ts` and have the limit error type and any upcoming error type that might arise in the future

rem1niscence (changes requested):

nymd (approved):

rem1niscence (approved):

PR #176, opened 2021-07-22
Incoming second layer of refactors for tests

rem1niscence:
- [x] Modify env loading strategy for `application.ts`  using observable hooks to have the option of  dynamically loading them at runtime 
- [x] Add package for mocking dependencies at runtime and apply mock for redis to be able to use it while testing while preserving functionality (this specially for mocking pocketjs on the acceptance tests later)
- [x] Add memory datasource for mocking repository transactions/db operations
- [x] Add blockchain controller acceptance tests
- [x] Small cleanup of warnings
- [x]  Set up coverage check


src/application.ts
```diff
@@ -104,7 +124,7 @@ export class PocketGatewayApplication extends BootMixin(ServiceMixin(RepositoryM
       50,
       100000,
       0,
-      120000,
+      1200000,
       false,
```
Evalir: Should we move these variables to the top and in uppercase, as they're constants?

src/datasources/gateway.datasource.ts
```diff
@@ -2,6 +2,10 @@ import { inject, lifeCycleObserver, LifeCycleObserver } from '@loopback/core'
 import { juggler } from '@loopback/repository'
 import { HttpErrors } from '@loopback/rest'
 
+// FIXME: on testing, the environment variable has to be set manually to a dummy
+// due to the loopback loading data sources loaded before any code from us, so
```
Evalir:
```suggestion
// due to loopback loading data sources loaded before any code from us, so
```

src/index.ts
```diff
@@ -6,8 +6,14 @@ const logger = require('./services/logger')
 export { PocketGatewayApplication }
 
 export async function main(options: ApplicationConfig = {}): Promise<PocketGatewayApplication> {
+  options.env = { load: true }
```
Evalir: Can't we just assign this by default on the method definition and set it as a constant at the top of the file too? something like
```js
const DEFAULT_APPLICATION_OPTIONS = { load: true }
```

tests/acceptance/blockchain.controller.acceptance.tests.ts
```diff
@@ -0,0 +1,120 @@
...
+
+    const res = await client.get('/blockchains').expect(200)
+
+    expect(res.body).to.be.Array()
+    expect(res.body).to.have.length(1)
+    expect(res.body[0]).to.containEql(expected)
+  })
+
+  it('retrieves the total count of blockchains', async () => {
+    await generateBlockchains(5)
```
Evalir, line 56: This method has side-effects, no? If so, I think we should at least return a boolean that the operation was successful and if not fail the test (because right now it feels like if it fails, it will just throw).

tests/acceptance/test-helper.ts
```diff
@@ -26,3 +35,38 @@ export interface AppWithClient {
   app: PocketGatewayApplication
   client: Client
 }
+
+const localEnv = {
```
Evalir: Another nitpick, but can we move this to the top of the file and name it something like `DUMMY_LOCAL_ENV`? We're referencing this before it was defined, and it's better form in TS to have the "constant" declarations at the top of the file :).

tests/fixtures/test.datasource.ts
```diff
@@ -0,0 +1,22 @@
+import { inject, lifeCycleObserver, LifeCycleObserver } from '@loopback/core'
+import { juggler } from '@loopback/repository'
+
+const config = {
```
Evalir, line 4: Can we uppercase this? :)

tests/acceptance/test-helper.ts
```diff
@@ -1,21 +1,30 @@
-import { PocketGatewayApplication } from '../..'
+import { PocketGatewayApplication } from '../../src/application'
 import { createRestAppClient, givenHttpServerConfig, Client } from '@loopback/testlab'
+import { gatewayTestDB } from '../fixtures/test.datasource'
+import RedisMock from 'ioredis-mock'
+import rewiremock from 'rewiremock'
```
Evalir: While this is super nitpicky, I feel over time this is important for code quality. Could we order the imports affected by this PR in a certain way? I'd go with:
- Library imports (`ioredis-mock`, etc) without `@` first
- Library imports with `@` after (loopback)
- Then relative imports.
  - I also like to alphabetize things, so I usually alphabetically order these as well.

And, I really like to alphabetize things, so, while I think you've actually been doing, we could also alphabetize the actual imports if they're destructured, like in line 2. :smile:

tests/acceptance/blockchain.controller.acceptance.tests.ts
```diff
@@ -0,0 +1,120 @@
...
+    const res = await client.get('/blockchains/1234').expect(200)
+
+    expect(res.body).to.be.Object()
+    expect(res.body).to.containEql(expected)
+
+    // Blockchain that doesn't exist
+    await client.get('/blockchains/nope').expect(404)
+  })
+
+  async function generateBlockchains(amount: number): Promise<Partial<Blockchains>> {
```
Evalir: Could we move these test helpers to the top of the file?

tests/acceptance/test-helper.ts
Evalir: And thinking about this, we could probably invert the two first rules to have loopback specific imports first, as this is a loopback project, just as its common to have express imports on top or React imports, for that matter.

Evalir (commented):
Mostly nitpicking, tests are looking good but haven't run them myself:

tests/acceptance/test-helper.ts
rem1niscence: I do certainly agree with that, we can also add this to the `CONTRIBUTING.md` file

src/application.ts
nymd: Yes, please. And let's comment as to what they are.

src/application.ts
```diff
@@ -104,7 +124,7 @@ export class PocketGatewayApplication extends BootMixin(ServiceMixin(RepositoryM
       50,
       100000,
       0,
-      120000,
+      1200000,
```
nymd: This grew. This is now a 20 minute default timeout!  lol, back to previous value please.

tests/acceptance/test-helper.ts
nymd: Isn't this something linting can do automatically?

nymd (changes requested):

tests/acceptance/test-helper.ts
rem1niscence: yep, I don't think at the specific level @Evalir mentioned but at least one good enough, on it

src/index.ts
```diff
@@ -3,10 +3,12 @@ import { ApplicationConfig } from '@loopback/core'
 
 const logger = require('./services/logger')
 
+const DEFAULT_APPLICATION_OPTIONS = { load: true }
+
 export { PocketGatewayApplication }
 
 export async function main(options: ApplicationConfig = {}): Promise<PocketGatewayApplication> {
-  options.env = { load: true }
+  options.env = DEFAULT_APPLICATION_OPTIONS
```
Evalir: Haha this is okay, but I think we could go further; I meant assigning a _default value_ to the `options` parameter. :slightly_smiling_face:

tests/acceptance/blockchain.controller.acceptance.tests.ts
```diff
@@ -1,24 +1,26 @@
-import { PocketGatewayApplication } from '../..'
 import { Client, expect } from '@loopback/testlab'
-import { setupApplication } from './test-helper'
-import { gatewayTestDB } from '../fixtures/test.datasource'
-import { BlockchainsRepository } from '../../src/repositories/blockchains.repository'
 import { Blockchains } from '../../src/models/blockchains.model'
+import { BlockchainsRepository } from '../../src/repositories/blockchains.repository'
+import { gatewayTestDB } from '../fixtures/test.datasource'
+import { PocketGatewayApplication } from '../..'
+import { setupApplication } from './test-helper'
```
Evalir: Whoops, I meant alphabetizing the _import paths_, apart from the imports themselves haha

Evalir (approved):
Some more nitpicking, but this LGTM—let's wait for Alex to sign off this

tests/acceptance/test-helper.ts
Evalir: Yeah, this can probably be achieved to some degree with linting, but there'll still be some nitpicking to do afterward haha.

nymd (approved):

PR #204, opened 2021-08-10
Pocketjs mock improvements

rem1niscence:
- [X] can add custom responses for custom requests
- [X] can add continuous responses to request, for example calling a sync check three times can return only two valid nodes, varying the responses on each call to the same request 
- [X] improve naming of methods, variables and add jsdoc convention
- [X] Fix bug on class instance where methods didn't really work, remove unused code

### DIFF
--- a/tests/mocks/pocketjs.ts
+++ b/tests/mocks/pocketjs.ts
@@ -1,3 +1,4 @@
+/* eslint-disable no-prototype-builtins */
 /* eslint-disable @typescript-eslint/naming-convention */
 import { Mock, It } from 'moq.ts'
 import { DEFAULT_POCKET_CONFIG } from '../../src/config/pocket-config'
@@ -21,8 +22,6 @@ import {
   RelayProof,
   RequestHash,
 } from '@pokt-network/pocket-js'
-
-const MINIMUM_NODE_STAKE = 15150
 
 export const DEFAULT_NODES = [
   new Node(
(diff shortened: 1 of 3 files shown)


tests/mocks/pocketjs.ts
```diff
@@ -115,23 +115,24 @@ const DEFAULT_MOCK_VALUES = {
...
   rpcProvider: HttpRpcProvider
   configuration: Configuration
   nodes: Node[]
-  // relayRequest: string = '{"method":"eth_blockNumber","params":[],"id":1,"jsonrpc":"2.0"}'
-  // relayResponse: string = '{"id":1,"jsonrpc":"2.0","result":"0x1083d57"}'
   fail = false
   rpcMockError = new RpcError('500', 'Mock error')
-  relayRequest = '{"method":"eth_blockNumber","params":[],"id":1,"jsonrpc":"2.0"}'
-  relayResponse: string | RelayResponse | RpcError = '{"id":1,"jsonrpc":"2.0","result":"0x1083d57"}'
+  relayResponse: Record<string, MockRelayResponse | Array<MockRelayResponse>> = {
```
derrandz, line 128: to be able to support multiple mocks we can export these to some fixture files and make explicit how we load from fixtures and inject responses into mocks so everyone can instantiate mocks however they like.

derrandz (commented):
👍🏻 Good work right there my man, love it!

crisog (approved):

PR #211, opened 2021-08-17
feat: Checkers should run in parallel when possible

### LINKED ISSUE #159: Chain and sync check can probably run at the same time
Currently chain check runs first and then sync check. It may be possible to refactor them to share an inherited class and make them run at the same time.

crisog:
Closes #159. Builds on #174.


crisog:
This PR also fixes a few problems that were present in the Pocket Relayer unit tests.

A new blockchain called `eth-mainnet-string` was introduced, but there were no `DEFAULT_NODES` that supported it. So, the chainIDCheck would always fail.

Two tests that needed the chainIDCheck & syncCheck mock were not using it (not sure why they didn't fail before, perhaps some logic was wrong).

crisog:
Acceptance tests are failing due to the same reason, but I do not have a lot of context about those tests. This is the error I get in the logs when the acceptance tests fail: `Error: ChainID check failure; using fallbacks`. 
cc: @rem1niscence

crisog:
I noticed that `v1.controller.acceptance` has several places sending `eth_` calls to `'mainnet'` host which is not an EVM chain, those should be failing too?
```
const res = await client
      .post('/v1/sd9fj31d714kgos42e68f9gh')
      .send({ method: 'eth_blockNumber', params: [], id: 1, jsonrpc: '2.0' })
      .set('Accept', 'application/json')
      .set('host', 'mainnet')
      .expect(200)
```

src/services/pocket-relayer.ts
```diff
@@ -408,6 +408,9 @@ export class PocketRelayer {
     )
 
     if (pocketSession instanceof Session) {
+      let syncCheckPromise: Promise<Node[]>
+      let chainCheckPromise: Promise<Node[]>
```
rem1niscence, line 416: these two variable names seem to be reduntant, could be changed to `syncCheck` and `chainCheck`. But that's up to you tho

rem1niscence (approved):

nymd (approved):

Issue #206, opened 2021-08-12
Always include altruist in the sync check pool


nymd:
With new chains like 0028 Eth Trace, it is easy to get a fully bad session when a very few number of node runners run a large number of nodes.

A bad session can pass sync check if all of the nodes agree on the incorrect sync.

To prevent this from happening, the altruist should always be included in the sync check group to provide another source of truth.

PR #216, opened 2021-08-17
Sync check now always require altruists

rem1niscence:
When doing a sync check now it always compares against the altruist block so now the nodes  have to pass both the session node check (all nodes within the sync allowance of the highest session node's block) and the altruist block height check with the same allowance. Addresses #206


nymd:
I like this approach because it doesn't rely on the altruist... it is just a secondary comparison. There is always the possibility that the altruist is out of sync.

I don't like this approach because the altruist check isn't fired off synchronously meaning it is adding a slight bit of time to the cold start that we already have an issue with. BUT the altruist check should be very quick and not add much.

My suggested approach on this was to add the altruist to the pool of nodes so it got synchronously fired all at once and the 5 nodes reaching consensus would actually be 6 nodes with the altruist in the mix.

I'm OK with the approach you took here as long as you take a look at my suggested approach and don't think it is significantly better.

rem1niscence:
Honestly don't think it adds enough improvement to justify the changes, so for now we can go with this and then if we really need to squeeze as much performance as possible we can do this

nymd:
I know @crisog has already started on merging the sync checker and chain checker; maybe to avoid the eventual conflict, we should integrate this change into his branch instead?

rem1niscence:
Yep,agree on that. Let's do it

crisog (approved):

PR #218, opened 2021-08-18
Tests/minor improvmenents

rem1niscence:
Apply some miscellaneous improvements to the tests, add a parallel command for faster execution, clean up spaces, refactor pocketConfiguration method to avoid repetition and change chain semantics on dummy data to be more real-world like.

Also add for  sync checker and chain checker log assertions, that is for behaviors that can only be asserted using logs to not only rely on coverage


nymd (changes requested):
I think this needs cleanup before merge as the new CI/linting will stop you for those warnings.

rem1niscence:
Fixed, remove warning, ready to go again

PR #229, opened 2021-08-23
[WIP] Filter nodes in session serving from blacklisted urls

### LINKED ISSUE #198: Filter nodes by service URL
It should not be possible to stake your node URL with a free tier end point.

It should not be possible to stake your node URL with a portal pocket endpoint, no matter the tier.

Node service URL should be analyzed before the node is selected for service and if it matches a list of blacklisted service URLs, it should be excluded from the session.

Suggested starting blacklist:

- infura
- alchemy
- portal.pokt

I want to exclude quicknode too since their free tier is slow but you can pay for a server there, too.

crisog:
Closes #198.

### DIFF
--- a/src/services/pocket-relayer.ts
+++ b/src/services/pocket-relayer.ts
@@ -496,6 +496,8 @@ export class PocketRelayer {
           return new Error('Sync / chain check failure; using fallbacks')
         }
       }
+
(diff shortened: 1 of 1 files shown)


crisog:
Actually, this is not possible.

PR #257, opened 2021-09-08
Improvements on logging

### LINKED ISSUE #255: Change influx bucket staging to which write
Change staging to write to mainnetRelayStaging as the bucket

```
ID			Name			Retention	Shard group duration	Organization ID
4aa0c7ae9525f28c	mainnetRelayStaging	infinite	n/a			a7d8bef207d2bd64
```

crisog:
Now fields won't show as undefined in the logs. Also, cleared up some variable names that were causing confusion.

For instance, `blockchain` in the database was referring to the string name f.e. `eth-mainnet`. However, in code it this same name was being used to store both `blockchainID` f.e `0021` and `chainID` f.e `100`. This PR also fixes some unit tests that were written while the variables were a bit misleading.


src/services/metrics-recorder.ts
```diff
@@ -136,29 +134,30 @@ export class MetricsRecorder {
...
-        blockchain,
+        blockchainID,
         serviceNode,
         elapsedTime,
         bytes,
         method,
         error,
       ]
 
+      console.log('Influx Blockchain Tag:', blockchainID)
```
nymd: Let's remove this since we are logging to a separate bucket and can see it there and so we won't have to do another commit to get rid of it later.

nymd (changes requested):

nymd (approved):

PR #268, opened 2021-09-10
Feat: Cache overruning session nodes stake

### LINKED ISSUE #262: Improve request flow for apps overrunning their stake
We could improve the API's response time for apps overrunning their stake if we better managed nodes that have exhausted their relays for that session. They shouldn't be marked as failures but some other sort of marking so that requests don't take 2x the time.

The current flow for an app that is over stake limit is:

- request sent to API
- API gets session and sends request to node
- node answers, sorry that app is over limit with me
- API then sends request to backup
- backup answers
- response returned

I think it would need a stateful session manager in the portal API to know the concept of a session and mark nodes out of it when it gets back the error message that the relays are exhausted for that node.

Example log:
`[2021-09-08T21:44:23.630Z] [[31merror[39m] [hJw9JFvupJ] [APP] [612d48ab5a586900341a302d] [9aa5c1c924b3acd4e6f15743c768cfc63e9e66bb90c3ead477c25cbe4f71733b] [the evidence is sealed, either max relays reached or claim already submitted] [0.341777761] [137] [https://www.xpollinate.io/] FAILURE`

rem1niscence:
Closes #262. This adds another cache layer to session nodes, now, when a session is created/obtained, a cache is created for ~~the maximum time of the session as set on the pocket configuration~~ two hours. From there on, if within a session a node fails due to overruning their max stake, it will be added to the session's cache and won't be include in future checks (sync and chain check), if a session has no nodes left, it will directly go to the altruists.

This change adds some important changes to take into consideration:

1. On sync check, when penalizing nodes out of sync, is based on the number of nodes available, not 5 nodes as it previously was
2. On chain check, when penalizing nodes reporting incorrect data, is based on the number of nodes available, not 5 nodes as it previously was
3. ~~Sync check only occurs now if there's at least 3 nodes available~~. No longer true

Besides that, no other major change was added, but feel free to comment on desired/expected behaviors from this.

### DIFF
--- a/src/services/chain-checker.ts
+++ b/src/services/chain-checker.ts
@@ -3,6 +3,8 @@ import { MetricsRecorder } from '../services/metrics-recorder'
 import { Redis } from 'ioredis'
 import { blockHexToDecimal, checkEnforcementJSON } from '../utils'
 import { createHash } from 'crypto'
+import { MAX_RELAYS_ERROR } from '../errors/types'
+import { removeNodeFromSession } from '../utils/cache'
 
 const logger = require('../services/logger')
 
@@ -26,6 +28,7 @@ export class ChainChecker {
     applicationPublicKey,
     pocketAAT,
     pocketConfiguration,
+    sessionKey,
   }: ChainIDFilterOptions): Promise<Node[]> {
     const CheckedNodes: Node[] = []
     let CheckedNodesList: string[] = []
@@ -79,6 +82,7 @@ export class ChainChecker {
       pocket,
       pocketAAT,
       pocketConfiguration,
+      sessionKey,
     }
     const nodeChainLogs = await this.getNodeChainLogs(options)
 
@@ -142,12 +146,12 @@ export class ChainChecker {
 
     // If one or more nodes of this session are not in Chain, fire a consensus relay with the same check.
     // This will penalize the out-of-Chain nodes and cause them to get slashed for reporting incorrect data.
-    if (CheckedNodes.length < 5) {
+    if (CheckedNodes.length < nodes.length) {
       const consensusResponse = await pocket.sendRelay(
         chainCheck,
         blockchainID,
         pocketAAT,
-        this.updateConfigurationConsensus(pocketConfiguration),
+        this.updateConfigurationConsensus(pocketConfiguration, nodes.length),
         undefined,
         'POST' as HTTPMethod,
         undefined,
@@ -179,6 +183,7 @@ export class ChainChecker {
     pocket,
     pocketAAT,
     pocketConfiguration,
+    sessionKey,
   }: GetNodesChainLogsOptions): Promise<NodeChainLog[]> {
     const nodeChainLogs: NodeChainLog[] = []
     const promiseStack: Promise<NodeChainLog>[] = []
@@ -203,6 +208,7 @@ export class ChainChecker {
         pocket,
         pocketAAT,
         pocketConfiguration,
+        sessionKey,
       }
 
       promiseStack.push(this.getNodeChainLog(options))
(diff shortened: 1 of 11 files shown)


src/services/pocket-relayer.ts
```diff
@@ -440,6 +441,24 @@ export class PocketRelayer {
       let nodes: Node[] = pocketSession.sessionNodes
       const relayStart = process.hrtime()
 
+      const cachedRemovedSessionNodes = await this.redis.get(`session-${pocketSession.sessionKey}`)
```
nymd: This is a good trick. I forgot pocket-js already had the concept of a sessionKey.

src/services/pocket-relayer.ts
```diff
@@ -440,6 +441,24 @@ export class PocketRelayer {
...
+
+      if (cachedRemovedSessionNodes) {
+        const nodesToRemove: string[] = JSON.parse(cachedRemovedSessionNodes)
+
+        nodes = nodes.filter((n) => !nodesToRemove.includes(n.publicKey))
+      } else {
+        // Maximum time in milliseconds for the next session to be rolloved,
```
nymd: * to be rolled over

I was a bit concerned about using the session length for anything given that block timing is variable. Optimal block timing is 15 minutes but sometimes they take up to 18 minutes.

I'm not sure we need to set this TTL with any sort of specificity. What's the harm if we set it for 2 hours?  It won't be pulled up if the sessionKey changes.

src/services/sync-checker.ts
```diff
@@ -263,12 +267,12 @@ export class SyncChecker {
...
     // If one or more nodes of this session are not in sync, fire a consensus relay with the same check.
     // This will penalize the out-of-sync nodes and cause them to get slashed for reporting incorrect data.
-    if (syncedNodes.length < 5) {
+    if (syncedNodes.length < nodes.length) {
       const consensusResponse = await pocket.sendRelay(
         syncCheck,
         blockchainID,
         pocketAAT,
-        this.updateConfigurationConsensus(pocketConfiguration),
+        this.updateConfigurationConsensus(pocketConfiguration, nodes.length),
```
nymd: This should not change the consensus node count. As you are sending a completely new relay and not specifying what node to send it to, PocketJS will use its own interpretation of what the session is and that includes all 5 nodes.

src/services/sync-checker.ts
```diff
@@ -512,11 +523,11 @@ export class SyncChecker {
     return nodeSyncLog
   }
 
-  updateConfigurationConsensus(pocketConfiguration: Configuration): Configuration {
+  updateConfigurationConsensus(pocketConfiguration: Configuration, consensusNodeCount: number): Configuration {
     return new Configuration(
       pocketConfiguration.maxDispatchers,
       pocketConfiguration.maxSessions,
-      5,
+      consensusNodeCount,
```
nymd: So this should go back to 5.

src/services/chain-checker.ts
```diff
@@ -346,11 +357,11 @@ export class ChainChecker {
     return nodeChainLog
   }
 
-  updateConfigurationConsensus(pocketConfiguration: Configuration): Configuration {
+  updateConfigurationConsensus(pocketConfiguration: Configuration, consensusNodeCount: number): Configuration {
     return new Configuration(
       pocketConfiguration.maxDispatchers,
       pocketConfiguration.maxSessions,
-      5,
+      consensusNodeCount,
```
nymd: Same as in other check, this should remain 5.

nymd (changes requested):

nymd (approved):

crisog (approved):

nymd (approved):

PR #286, opened 2021-09-19
Refactor session nodes cache

rem1niscence:
Currently, the session nodes cache was handled by a redis string key and several operations around that, this refactor leverages the data type used to a redis set to avoid computing the existence of keys on server code, removes the need of a queue on setting the key value and takes advantage of the set data type to always guarantee that node public keys won't be added more than once


crisog (approved):
Nice improvement! These methods from `io-redis` are definitely the best way to handle it.

nymd (approved):

PR #288, opened 2021-09-23
feat: return error when relay method requires WS

### LINKED ISSUE #287: Parse incoming relay requests for calls that require websockets
Using the same method of relay inspection that we use for the getLogs limits, we should add a custom error when users send in calls that require websockets.  The context is that Infura allows these calls over HTTPS and internally routes them to websockets so devs are accustomed to this behaviour.

> eth_newFilter
> eth_getFilterChanges
> others?

The relay request should be immediately returned with the error and never sent to the network or altruists.

crisog:
Closes #287.

### DIFF
--- a/src/services/pocket-relayer.ts
+++ b/src/services/pocket-relayer.ts
@@ -19,6 +19,14 @@ const logger = require('../services/logger')
 import axios from 'axios'
 import { removeNodeFromSession } from '../utils/cache'
 
+const WS_ONLY_METHODS = [
+  'eth_subscribe',
+  'eth_newFilter',
+  'newBlockFilter',
+  'eth_getFilterChanges',
+  'eth_getFilterLogs',
+]
+
 export class PocketRelayer {
   host: string
   origin: string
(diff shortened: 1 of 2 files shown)


src/services/pocket-relayer.ts
```diff
@@ -773,7 +781,11 @@ export class PocketRelayer {
     blockchainID: string,
     logLimitBlocks: number
   ): Promise<string | Error> {
-    if (parsedRawData.method === 'eth_getLogs') {
+    if (WS_ONLY_METHODS.includes(parsedRawData.method)) {
+      return new HttpErrors.BadRequest(
+        `We cannot serve ${parsedRawData.method} method over HTTPS. At the moment, we do not support WebSockets.`
```
rem1niscence, line 786: I would only change the message to `${parsedRawData.method} method cannot be served over HTTPS. WebSockets are not supported  at the moment.`

rem1niscence (approved):

PR #289, opened 2021-09-23
feat: log final elapsed time in milliseconds

rem1niscence:
Does not affect influx metrics calculation, only log

### DIFF
--- a/src/services/metrics-recorder.ts
+++ b/src/services/metrics-recorder.ts
@@ -86,9 +86,11 @@ export class MetricsRecorder {
   }): Promise<void> {
     try {
       let elapsedTime = 0
+      let elapsedTimeMs = 0
       const relayEnd = process.hrtime(relayStart)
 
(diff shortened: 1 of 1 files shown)


src/services/metrics-recorder.ts
```diff
@@ -86,9 +86,11 @@ export class MetricsRecorder {
   }): Promise<void> {
     try {
       let elapsedTime = 0
+      let elapsedTimeMs = 0
       const relayEnd = process.hrtime(relayStart)
 
       elapsedTime = (relayEnd[0] * 1e9 + relayEnd[1]) / 1e9
+      elapsedTimeMs = Math.ceil((relayEnd[0] * 1e9 + relayEnd[1]) / 1e6)
```
crisog, line 93: This one was really tough to review.. :kek: Looks good though!

I think I would have switched the term, put the milliseconds at `elapsedTime` and then `elapsedTimeSec` cause it feels more natural but either is fine.

crisog (approved):

PR #290, opened 2021-09-23
Reset cache on exhaust

rem1niscence:
Once a node is removed, there's a period where it still can be taken on the session, leading to an expected failure following an altruist relay, which is an undesired overheat. This ensures that now everytime a node is removed, upcoming requests will need to do a sync check/chain check to reverify the behavior of remaining nodes


rem1niscence:
Another behavior could be taking the node out of the sync/chain check caches and update it again but I'd rather just retrigger the checks again so we can verify how the remaining nodes are doing, lmk if you think otherwise

nymd (approved):

crisog (approved):

PR #299, opened 2021-10-04
feat: non-evm chains sync check

crisog:
This PR will allow us to perform sync checks to any chain, not only EVM-based chains.

### DIFF
--- a/src/services/sync-checker.ts
+++ b/src/services/sync-checker.ts
@@ -533,7 +533,7 @@ export class SyncChecker {
     )
   }
 
-  parseBlockFromPayload(payload: any, syncCheckResultKey: string): number {
+  parseBlockFromPayload(payload: object, syncCheckResultKey: string): number {
     const rawHeight = payload[`${syncCheckResultKey}`]
     const blockHeight = isBlockHex(rawHeight) ? blockHexToDecimal(rawHeight) : parseInt(rawHeight)
 
--- a/src/services/sync-checker.ts
+++ b/src/services/sync-checker.ts
@@ -1,7 +1,7 @@
 import { Configuration, HTTPMethod, Node, Pocket, PocketAAT, RelayResponse } from '@pokt-network/pocket-js'
 import { MetricsRecorder } from '../services/metrics-recorder'
 import { Redis } from 'ioredis'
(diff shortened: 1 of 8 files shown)


src/services/sync-checker.ts
```diff
@@ -539,6 +532,13 @@ export class SyncChecker {
       pocketConfiguration.rejectSelfSignedCertificates
     )
   }
+
+  parseBlockFromPayload(payload: any, syncCheckResultKey: string): number {
+    const rawHeight = payload[`${syncCheckResultKey}`]
```
rem1niscence, line 537: This does not work with chains which result is in a nested object, right now the current chains does not have the issue, but could add a TODO atm for when's needed

src/services/sync-checker.ts
```diff
@@ -394,24 +390,21 @@ export class SyncChecker {
     const relayStart = process.hrtime()
 
     const relayResponse = await pocket.sendRelay(
-      syncCheck,
+      blockchainSyncCheck.body,
```
rem1niscence, line 393: Might come as undefined, plz add a default string value

src/services/sync-checker.ts
```diff
@@ -394,24 +390,21 @@ export class SyncChecker {
...
       blockchainID,
       pocketAAT,
       this.updateConfigurationTimeout(pocketConfiguration),
       undefined,
       'POST' as HTTPMethod,
-      syncCheckPath,
+      blockchainSyncCheck.path,
```
rem1niscence: Might come as undefined, plz add a default string value

rem1niscence (changes requested):

src/services/sync-checker.ts
```diff
@@ -539,6 +532,13 @@ export class SyncChecker {
       pocketConfiguration.rejectSelfSignedCertificates
     )
   }
+
+  parseBlockFromPayload(payload: object, syncCheckResultKey: string): number {
+    const rawHeight = payload[`${syncCheckResultKey}`]
```
rem1niscence: This does not work with chains which result is in a nested object, right now the current chains does not have the issue, but could add a TODO atm for when's needed

rem1niscence (changes requested):

rem1niscence (approved):

PR #327, opened 2021-10-11
Feat: Add redis prefix key

rem1niscence:
Adds a redis prefix key to be able to rotate the keys on each deploy

### DIFF
--- a/src/application.ts
+++ b/src/application.ts
@@ -64,6 +64,7 @@ export class PocketGatewayApplication extends BootMixin(ServiceMixin(RepositoryM
       DEFAULT_LOG_LIMIT_BLOCKS,
       AAT_PLAN,
       REDIRECTS,
+      COMMIT_HASH,
     } = await this.get('configuration.environment.values')
 
     const environment: string = NODE_ENV || 'production'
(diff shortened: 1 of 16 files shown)


crisog (approved):

src/application.ts
```diff
@@ -79,6 +80,7 @@ export class PocketGatewayApplication extends BootMixin(ServiceMixin(RepositoryM
     const defaultLogLimitBlocks: number = parseInt(DEFAULT_LOG_LIMIT_BLOCKS) || 10000
     const aatPlan = AAT_PLAN || AatPlans.PREMIUM
     const redirects: string | object[] = REDIRECTS || ''
+    const commitHash: string | string = COMMIT_HASH || ''
```
nymd, line 83: If for some reason the commit hash fails, what about just replacing it with a quick random string instead of blank. Since this is only set when the application boots, it would have the same basic effect.

nymd (changes requested):

src/application.ts
rem1niscence: I dont think so, on that case each instance will have its own cache, as the hash will be calculated on runtime and will result in the same values repeated times the maximum amount of instances, slowing the gateway specially on sync/chain checks

nymd: Hm true.

PR #328, opened 2021-10-11
fix: always altruist when missing chain check

crisog:
This logic was modified to solve a known bug, during a small refactor we missed a logic error while reviewing the code. This bug would cause all chains without chain check to go to altruists.

### DIFF
--- a/src/services/pocket-relayer.ts
+++ b/src/services/pocket-relayer.ts
@@ -520,6 +520,26 @@ export class PocketRelayer {
           chainCheckedNodes = (chainCheckResult as PromiseFulfilledResult<Node[]>).value
         }
       } else {
+        const error = 'ChainID check failure'
+        const method = 'checks'
+
+        await this.metricsRecorder.recordMetric({
+          requestID,
+          applicationID: application.id,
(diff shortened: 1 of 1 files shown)


src/services/pocket-relayer.ts
```diff
@@ -515,14 +515,18 @@ export class PocketRelayer {
 
       const [chainCheckResult, syncCheckResult] = await checkersPromise
 
-      if (blockchainIDCheck && this.isCheckPromiseResolved(chainCheckResult)) {
-        chainCheckedNodes = (chainCheckResult as PromiseFulfilledResult<Node[]>).value
+      if (blockchainIDCheck) {
+        if (this.isCheckPromiseResolved(chainCheckResult)) {
+          chainCheckedNodes = (chainCheckResult as PromiseFulfilledResult<Node[]>).value
+        }
       } else {
```
nymd: Won't this always fail if no chain ID check is set?

nymd (changes requested):

src/services/pocket-relayer.ts
crisog: Yea, the else goes along the promise valuation.

nymd (approved):

Issue #111, opened 2021-07-07
Inconsistent response codes


derrandz:
```
  // The gateway constantly responds with 200
  // and responds with errors in response.data
  // in a non-consistent form.
  // So can't rely on response code
 ```

nymd:
If the error in the response data is in a properly formatted JSON response then the API is technically doing it's job by returning the relay. There is code in the API to catch these properly formatted JSON errors:

https://github.com/pokt-foundation/portal-api/blob/4dd8faa27550e3ee3af90c7903a6d8a4d52be87f/src/services/pocket-relayer.ts#L430

It is hacky. If it is not catching this case for this issue, can you give a full example of the request/response?

derrandz:
Indeed, that's what an accepted behavior of a relayer endpoint should look like. I must've lost context as I was expecting errors whereas now --since it's a relay, they will be part of the data relayed.

This is now to be reversed, this is an issue of the error handling code at the datasource in the wallet and explorer and not the portal-api.

I will close this issue in favor of the other two when I created them.

- [ ] Reference this issue and create its reverse in wallet
- [ ] Do the same for the explorer
- [ ] Close this issue.

rem1niscence:
Are we sure we want all the responses to be checked on the body and not some on the response code? for example as of right now relaying a non existent app will return 200 but a message with "Entity not found" which could be semantically misleading and the relay didn't even happen

PR #331, opened 2021-10-11
ref: polishing pocket relayer

### LINKED ISSUE #329: Refactor Pocket Relayer
It's almost a thousand lines of code. There's a lot of code that could be either generalized, or moved elsewhere such as `checkSecretKey`, `checkWhitelist`, a few interfaces, etc.

crisog:
Closes #329.

Moved a lot of code to `utils/` so we can reuse it later, and not use it from the pocket relayer instance which is not right cause nothing to do with relayer, a lot of utility functions actually.

Also, a changed interfaces to types since these are not really being used as interfaces. There's still a lot to improve.


src/controllers/v1.controller.ts
```diff
@@ -8,9 +8,11 @@ import { Redis } from 'ioredis'
 import { Pool as PGPool } from 'pg'
 import { CherryPicker } from '../services/cherry-picker'
 import { MetricsRecorder } from '../services/metrics-recorder'
-import { PocketRelayer, SendRelayOptions } from '../services/pocket-relayer'
+import { PocketRelayer } from '../services/pocket-relayer'
 import { SyncChecker } from '../services/sync-checker'
 import { ChainChecker } from '../services/chain-checker'
+import { SendRelayOptions } from '../utils/types'
+import { loadBlockchain } from '../utils/relayer'
```
Evalir: Alrighty, if you requested my review, be prepared for the comments :)...

Let's reorder these imports alphabetically by relative path.

src/controllers/v1.controller.ts
```diff
@@ -157,7 +159,15 @@ export class V1Controller {
           blockchainEnforceResult: _enforceResult,
           // eslint-disable-next-line
           blockchainSyncCheck: _syncCheck,
-        } = await this.pocketRelayer.loadBlockchain()
+        } = await loadBlockchain(this.host, this.redis, this.blockchainsRepository, this.defaultLogLimitBlocks).catch(
```
Evalir: Let's avoid using this catch—we already have a `try` clause here. I think it's quite big so, if we can break it up into steps, that'd be great.

Evalir: Also, this isn't letting me comment on unaffected code, but, this whole logic is inside an `if`, line 155. I think it'd be easier if we checked for the reverse case: if there's *no* LB ID, we can just return (and we should actually throw an error here probably).

src/controllers/v1.controller.ts
```diff
@@ -157,7 +159,15 @@ export class V1Controller {
...
+            logger.log('error', `Incorrect blockchain: ${this.host}`, {
+              origin: this.origin,
+            })
+            throw new HttpErrors.BadRequest(`Incorrect blockchain: ${this.host}`)
+          }
+        )
+
         // Fetch applications contained in this Load Balancer. Verify they exist and choose
         // one randomly for the relay.
         const application = await this.fetchLoadBalancerApplication(
```
Evalir: Same comment of inverting the `if` here with the `application.id` check—if there's none, we can return or throw an error.

My logic for doing this, is that we'd usually like to make sure that the happy path isn't contained inside a condition—for someone inspecting the code from the outside, There's less mental overhead with thinking that things won't happen if  `x` and `y` and `z` (and so on) are not there, rather than having to keep track that _if they exist_, things will happen.

src/services/chain-checker.ts
```diff
@@ -1,9 +1,10 @@
 import { Configuration, HTTPMethod, Node, Pocket, PocketAAT, RelayResponse } from '@pokt-network/pocket-js'
 import { MetricsRecorder } from '../services/metrics-recorder'
 import { Redis } from 'ioredis'
-import { blockHexToDecimal, checkEnforcementJSON, getNodeNetworkData } from '../utils'
-import { MAX_RELAYS_ERROR } from '../errors/types'
-import { removeNodeFromSession } from '../utils/cache'
+import { checkEnforcementJSON } from '../utils/string'
+import { getNodeNetworkData, removeNodeFromSession } from '../utils/cache'
+import { blockHexToDecimal } from '../utils/block'
+import { MAX_RELAYS_ERROR } from '../utils/constants'
```
Evalir: Let's also reorder these.

src/services/limiter.ts
```diff
@@ -0,0 +1,82 @@
+import { JSONObject } from '@loopback/context'
+import { HttpErrors } from '@loopback/rest'
+import axios from 'axios'
+import { LimitError } from '../errors/types'
+import { blockHexToDecimal } from '../utils/block'
+
```
Evalir: No need for this extra line here.

src/services/limiter.ts
```diff
@@ -0,0 +1,82 @@
...
+export async function enforceEVMLimits(
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  parsedRawData: Record<string, any>,
+  blockchainID: string,
+  logLimitBlocks: number,
+  altruists: JSONObject
+): Promise<void | Error> {
+  if (WS_ONLY_METHODS.includes(parsedRawData.method)) {
+    return new HttpErrors.BadRequest(
+      `We cannot serve ${parsedRawData.method} method over HTTPS. At the moment, we do not support WebSockets.`
```
Evalir:
```suggestion
      `We cannot serve the ${parsedRawData.method} method over HTTPS. At the moment, we do not support WebSockets.`
```

src/services/limiter.ts
```diff
@@ -0,0 +1,82 @@
...
+    )
+  } else if (parsedRawData.method === 'eth_getLogs') {
+    let toBlock: number
+    let fromBlock: number
+    let isToBlockHex = false
+    let isFromBlockHex = false
+    const altruistUrl = String(altruists[blockchainID])
```
Evalir, line 26: What if there's no `altruistUrl`? I think we should maybe throw an error here unless we intentionally don't use it if there's none?

src/services/limiter.ts
```diff
@@ -0,0 +1,82 @@
...
+      fromBlock = blockHexToDecimal(fromBlockParam)
+      isFromBlockHex = true
+    }
+
+    if ((toBlock !== 0 || fromBlock !== 0) && altruistUrl !== 'undefined') {
+      // Altruist
+      // TODO: use a generic getHeightFromAltruist function to fetch altruist block height
+      const rawData = JSON.stringify({ jsonrpc: '2.0', id: 1, method: 'eth_blockNumber', params: [] })
+
+      let axiosConfig = {}
```
Evalir: Let's declare this object inside the `try` and put it as `const`. Objects are mutable even if they're constants—but will let us avoid changing the _type_ of the object, which is what we want.

src/services/limiter.ts
```diff
@@ -0,0 +1,82 @@
...
+
+        if (!isToBlockHex) {
+          toBlock = latestBlock
+        }
+        if (!isFromBlockHex) {
+          fromBlock = latestBlock
+        }
+      } catch (e) {
+        logger.log('error', `Failed trying to reach altruist (${altruistUrl}) to fetch block number.`)
+        return new HttpErrors.InternalServerError('Internal error. Try again with a explicit block number.')
```
Evalir, line 65: Not very sure how we would solve this haha... But the error logged and the error reported back are very different and misleading.

src/services/limiter.ts
```diff
@@ -0,0 +1,82 @@
...
+      }
+    } else {
+      // We cannot move forward if there is no altruist available.
+      if (!isToBlockHex || !isFromBlockHex) {
+        return new LimitError(`Please use an explicit block number instead of 'latest'.`, parsedRawData.method)
+      }
```
Evalir, line 71: How does this comment correspond to the code below?

src/services/limiter.ts
```diff
@@ -0,0 +1,82 @@
...
+    }
+    if (toBlock - fromBlock > logLimitBlocks) {
```
Evalir, line 73: We don't need to change this, but this is a particular pet peeve of mine when it comes to using `if` statements. I tend to like just using `if`s and not `else if`s because a lot of times we're expecting the logic to "fall through", just like a switch statement. If we were to change it, I'd remove the `else if` and the `else` we have above.

src/services/sync-checker.ts
```diff
@@ -1,13 +1,14 @@
 import { Configuration, HTTPMethod, Node, Pocket, PocketAAT, RelayResponse } from '@pokt-network/pocket-js'
 import { MetricsRecorder } from '../services/metrics-recorder'
 import { Redis } from 'ioredis'
-import { blockHexToDecimal, checkEnforcementJSON, getNodeNetworkData } from '../utils'
+import { removeNodeFromSession, getNodeNetworkData } from '../utils/cache'
+import { checkEnforcementJSON } from '../utils/string'
+import { blockHexToDecimal } from '../utils/block'
```
Evalir: :)

src/services/sync-checker.ts
```diff
@@ -1,13 +1,14 @@
...
 
 const logger = require('../services/logger')
 
 import axios from 'axios'
```
Evalir: Let's move this up with the libray imports. Let's try to always use this order:

- `node_modules` imports (`axios`, `ioredis`, etc...)
- `node_modules` imports with a `@`
- relative imports.
- `require` imports because they're ugly.

src/utils/constants.ts
```diff
@@ -0,0 +1,9 @@
+export const WS_ONLY_METHODS = [
+  'eth_subscribe',
+  'eth_newFilter',
+  'newBlockFilter',
+  'eth_getFilterChanges',
+  'eth_getFilterLogs',
```
Evalir: Let's alphabetize these :)

src/utils/pocket.ts
```diff
@@ -0,0 +1,43 @@
+import { Configuration } from '@pokt-network/pocket-js'
+import { Decryptor } from 'strong-cryptor'
+import { Applications } from '../models'
```
Evalir: :)

src/utils/pocket.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  if (
+    application.gatewaySettings.secretKeyRequired && // If the secret key is required by app's settings // and
+    application.gatewaySettings.secretKey && // the app's secret key is set // and
+    (!secretKeyDetails.secretKey || // the request doesn't contain a secret key // or
+      secretKeyDetails.secretKey.length < 32 || // the secret key is invalid // or
+      (secretKeyDetails.secretKey.length === 32 &&
+        secretKeyDetails.secretKey !== application.gatewaySettings.secretKey) || // the secret key does not match plaintext // or
+      (secretKeyDetails.secretKey.length > 32 &&
+        secretKeyDetails.secretKey !== decryptor.decrypt(application.gatewaySettings.secretKey))) // does not match encrypted
```
Evalir: This logic is hard to read. Maybe we can put the individual parts in variables and then `OR` them together? Something like

```
const secretKeySet = secretKeyRequered && secretKey
const secretKeyChecks = ...
```

src/utils/pocket.ts
```diff
@@ -0,0 +1,43 @@
...
+  ) {
+    return false
+  }
+  return true
+}
+
+export type SecretKeyDetails = {
+  databaseEncryptionKey: string
+  secretKey: string
+}
```
Evalir: How is the repo ordering types? Do they go at the top of the page or at the bottom of the page? If there's no set style, we should pick one and start moving everything where it should go.

src/utils/relayer.ts
```diff
@@ -0,0 +1,107 @@
...
+import { BlockchainsRepository } from '../repositories'
+import { BlockchainDetails } from './types'
+
+// Fetch node client type if Ethereum based
+export async function fetchClientTypeLog(
+  redis: Redis,
+  blockchainID: string,
+  id: string | undefined
+): Promise<string | null> {
+  const clientTypeLog = await redis.get(blockchainID + '-' + id + '-clientType')
```
Evalir:
```suggestion
  const clientTypeLog = await redis.get(`${blockchainID}-${id}-clientType`)
```

src/utils/relayer.ts
```diff
@@ -0,0 +1,107 @@
...
+    blockchains = await blockchainsRepository.find()
+    await redis.set('blockchains', JSON.stringify(blockchains), 'EX', 60)
+  } else {
+    blockchains = JSON.parse(cachedBlockchains)
+  }
+
+  // Split off the first part of the request's host and check for matches
+  const [blockchainRequest] = host.split('.')
+
+  const blockchainFilter = blockchains.filter(
```
Evalir: Seeing that we're accessing the first array element only, we could probably do
```suggestion
  const [blockchainFilter] = blockchains.filter(
```

And that way we can just check below if it's defined (or if it's not and return, like I had suggested above in how to treat `if`s).

src/utils/relayer.ts
```diff
@@ -0,0 +1,107 @@
...
+    (b: { blockchain: string }) => b.blockchain.toLowerCase() === blockchainRequest.toLowerCase()
+  )
+
+  if (blockchainFilter[0]) {
+    let blockchainEnforceResult = ''
+    let blockchainIDCheck = ''
+    let blockchainID = ''
+    let blockchainChainID = ''
+    let blockchainLogLimitBlocks = 10000 // Should never be 0
```
Evalir: This `10000` should be a constant at the top of the file; something like

```
const DEFAULT_BLOCK_LOGS_LIMIT = 10000
```

src/utils/types.ts
```diff
@@ -0,0 +1,25 @@
...
+  rawData: object | string
+  relayPath: string
+  httpMethod: HTTPMethod
+  application: Applications
+  requestID: string
+  requestTimeOut?: number
+  overallTimeOut?: number
+  relayRetries?: number
+  logLimitBlocks?: number
+}
```
Evalir: Let's order all of these alphabetically.

tests/unit/chain-checker.unit.ts
```diff
@@ -8,7 +8,7 @@ import { CherryPicker } from '../../src/services/cherry-picker'
 import { MetricsRecorder } from '../../src/services/metrics-recorder'
 import { metricsRecorderMock } from '../mocks/metricsRecorder'
 import { DEFAULT_NODES, PocketMock } from '../mocks/pocketjs'
-import { MAX_RELAYS_ERROR } from '../../src/errors/types'
+import { MAX_RELAYS_ERROR } from '../../src/utils/constants'
 import MockAdapter from 'axios-mock-adapter'
```
Evalir: Let's remember to put these with the other library imports so we're not mixing them :).

tests/unit/pocket-relayer.unit.ts
```diff
@@ -17,7 +17,11 @@ import { metricsRecorderMock } from '../mocks/metricsRecorder'
 import { DEFAULT_NODES, PocketMock } from '../mocks/pocketjs'
 import { BlockchainsRepository } from '../../src/repositories/blockchains.repository'
 import { gatewayTestDB } from '../fixtures/test.datasource'
-import { LimitError, MAX_RELAYS_ERROR } from '../../src/errors/types'
+import { LimitError } from '../../src/errors/types'
+import { MAX_RELAYS_ERROR } from '../../src/utils/constants'
+import { checkWhitelist, parseMethod } from '../../src/utils/string'
+import { checkSecretKey, updateConfiguration } from '../../src/utils/pocket'
+import { loadBlockchain } from '../../src/utils/relayer'
```
Evalir: :)

Evalir (commented):
It's looking a bit cleaner. Let's take it a step further:

src/controllers/v1.controller.ts
crisog: Ordered alphabetically from normal ones, then ones with @, and lastly `../` imports.

src/controllers/v1.controller.ts
crisog: I agree :)

src/controllers/v1.controller.ts
crisog: Done; I also think this is a best approach. Thx!

src/services/limiter.ts
crisog: Correct, it is not used if its `undefined`. Ideally, we won't even fetch anything from the altruist here. There should be a function outside of `limiter.ts` called `getHeightFromAltruist()` that does the job instead of using axios here. Saving that for the next refactor round.

src/services/limiter.ts
crisog: Done :)

src/services/limiter.ts
crisog: This was on purpose. The HttpError returned is used in the client, and serves a different purpose than the logs. The logs are useful for engineers to debug, so I think a different message makes sense in such scenarios. 

The user making the request doesn't really need to know that we failed communicating with an altruist, they just need to know that we are unable to serve their request as-is, and they needs to use a explicit block for it to work.

src/services/limiter.ts
crisog: Same as above, this error goes back to the user that sent the request. They don't even know what an altruist is, they just needs instructions.

However, all this altruist logic should go out of the limiter anyway.

src/utils/pocket.ts
crisog: I'd say we mostly put them in the bottom. We'll use that, I'll check if its like that everywhere.

src/utils/pocket.ts
crisog: Done! 

This is a lot of logic so we need to make sure its working as expected after the refactor.

src/controllers/v1.controller.ts
```diff
@@ -150,59 +153,58 @@ export class V1Controller {
...
-        // Fetch applications contained in this Load Balancer. Verify they exist and choose
-        // one randomly for the relay.
-        const application = await this.fetchLoadBalancerApplication(
-          loadBalancer.id,
-          loadBalancer.applicationIDs,
-          blockchain,
-          filter
-        )
+      if (!loadBalancer?.id) {
+        throw new HttpErrors.InternalServerError('Load balancer configuration error')
```
rem1niscence: I know is refactor but this has been bothering me for a while, this is not a configuration error nor 500, but is a not found, not sure if should be fixed here but this has already been commented on #111. For now I'll suggest to change to not found and then let's see what to do with the response code

src/utils/constants.ts
```diff
@@ -0,0 +1,11 @@
+export const DEFAULT_BLOCK_LOGS_LIMIT = 10000 // Should never be 0
```
rem1niscence: As you can see is already with a default on the environment variables, this can be safely deleted

src/utils/relayer.ts
```diff
@@ -0,0 +1,110 @@
...
+  redis: Redis,
+  blockchainID: string,
+  id: string | undefined
+): Promise<string | null> {
+  const clientTypeLog = await redis.get(`${blockchainID}-${id}-clientType`)
+
+  return clientTypeLog
+}
+
+export function isCheckPromiseResolved(promise: PromiseSettledResult<Node[]>): boolean {
```
rem1niscence, line 20: this doesn't look like is part of the relayer, it can be moved to a miscelaneus utils file

src/utils/relayer.ts
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  // Split off the first part of the request's host and check for matches
+  const [blockchainRequest] = host.split('.')
+
+  const [blockchainFilter] = blockchains.filter(
+    (b: { blockchain: string }) => b.blockchain.toLowerCase() === blockchainRequest.toLowerCase()
+  )
+
+  if (blockchainFilter) {
```
rem1niscence: here comes I with my pet peeves, I'm a fan of having the main flow outside any block, that means the expected return value should not be anywhere else but at the end of the file, so when I read code I know that everything outside blocks is the "desired" behavior. So why not better change the `if` to check if not there and throw on that case and continue the flow normally on the main block.

src/utils/relayer.ts
```diff
@@ -0,0 +1,110 @@
...
+  const clientTypeLog = await redis.get(`${blockchainID}-${id}-clientType`)
+
+  return clientTypeLog
+}
+
+export function isCheckPromiseResolved(promise: PromiseSettledResult<Node[]>): boolean {
+  return promise.status === 'fulfilled' && promise.value !== undefined && promise.value.length > 0
+}
+
+export function filterCheckedNodes(syncCheckNodes: Node[], chainCheckedNodes: Node[]): Node[] {
```
rem1niscence, line 24: you can directly return here, the variable is not neccesary as I already know from the types that the fn must return  an array of nodes

src/utils/string.ts
```diff
@@ -0,0 +1,59 @@
+export function checkEnforcementJSON(test: string): boolean {
```
rem1niscence: file name is not intuitive, this can maybe be called `checkers` or even still have the name `enforcements` as is kinda what is happening here.

src/utils/string.ts
```diff
@@ -0,0 +1,59 @@
...
+  }
+  test = test.replace(/\\(?:["\\\/bfnrt]|u[0-9a-fA-F]{4})/g, '@')
+  test = test.replace(/"[^"\\\n\r]*"|true|false|null|-?\d+(?:\.\d*)?(?:[eE][+\-]?\d+)?/g, ']')
+  test = test.replace(/(?:^|:|,)(?:\s*\[)+/g, '')
+  return /^[\],:{}\s]*$/.test(test)
+}
+
+// Check passed in string against an array of whitelisted items
+// Type can be "explicit" or substring match
+export function checkWhitelist(tests: string[], check: string, type: string): boolean {
```
rem1niscence: also the checkWhilelist and secret key are on different files and I think they should be semantically together as kinda serve the same purpose, but up to you

src/utils/relayer.ts
crisog: In theory, nothing in this file is exclusive to the relayer; it's just being using there for now. Considering that, I don't think it's worth moving this one in particular outside and leaving the others there.

src/utils/relayer.ts
crisog: Definitely. Enrique pointed it out on his review; apparently I missed it on this place (perhaps in others too?). Thx!

src/utils/string.ts
crisog: Yea, actually there was a file with that name before. The rationale behind moving it to a generic name, was putting all of the string evaluation methods there, hence adding the parseMethod there too. At the end, `enforcements` is more specific and accurate to what these methods are doing. So I'll rename it accordingly - also, I will move `parseMethod` to a new file called `parsing.ts` so I can leave only the enforcements there.

rem1niscence (approved):

PR #336, opened 2021-10-18
Slash timeout archival

rem1niscence:
some nodes on archival are failing due to misconfigured timeouts on their end, this slashes them and removes from the session if fail continously

### DIFF
--- a/src/application.ts
+++ b/src/application.ts
@@ -62,6 +62,7 @@ export class PocketGatewayApplication extends BootMixin(ServiceMixin(RepositoryM
       AAT_PLAN,
       REDIRECTS,
       COMMIT_HASH,
+      ARCHIVAL_CHAINS,
     } = await this.get('configuration.environment.values')
 
     const environment: string = NODE_ENV || 'production'
@@ -77,7 +78,8 @@ export class PocketGatewayApplication extends BootMixin(ServiceMixin(RepositoryM
     const defaultLogLimitBlocks: number = parseInt(DEFAULT_LOG_LIMIT_BLOCKS) || 10000
     const aatPlan = AAT_PLAN || AatPlans.PREMIUM
     const redirects: string | object[] = REDIRECTS || ''
-    const commitHash: string | string = COMMIT_HASH || ''
+    const commitHash: string = COMMIT_HASH || ''
+    const archivalChains: string[] = (ARCHIVAL_CHAINS || []).split(',')
 
     if (!dispatchURL) {
       throw new HttpErrors.InternalServerError('DISPATCH_URL required in ENV')
@@ -200,5 +202,6 @@ export class PocketGatewayApplication extends BootMixin(ServiceMixin(RepositoryM
     const hash = crypto.createHash('md5').update(parts.join(''))
 
     this.bind('processUID').to(hash.digest('hex'))
+    this.bind('archivalChains').to(archivalChains)
   }
 }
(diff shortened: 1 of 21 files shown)


src/utils/constants.ts
```diff
@@ -9,3 +9,12 @@ export const WS_ONLY_METHODS = [
   'eth_subscribe',
   'newBlockFilter',
 ]
+
+export const ARCHIVAL_CHAINS = [
```
rem1niscence: should this be an env?

nymd: Yes please

src/services/cherry-picker.ts
```diff
@@ -1,9 +1,19 @@
...
 
 const logger = require('../services/logger')
 
+// Amount of times a node is allowed to fail due to misconfigured timeout before
+// being removed from the session
+const TIMEOUT_LIMIT = 20
+
+// Allowed difference on timeout, expressed in seconds, as the timeout usually
+// wont be exact
+const TIMEOUT_THRESHOLD = 2
```
nymd: maybe TIMEOUT_VARIANCE is a better name for this

src/services/cherry-picker.ts
```diff
@@ -1,9 +1,19 @@
...
 import { Node } from '@pokt-network/pocket-js'
 import { Applications } from '../models'
+import { getNodeNetworkData, removeNodeFromSession } from '../utils/cache'
+import { ARCHIVAL_CHAINS } from '../utils/constants'
 
 const logger = require('../services/logger')
 
+// Amount of times a node is allowed to fail due to misconfigured timeout before
+// being removed from the session
+const TIMEOUT_LIMIT = 20
```
rem1niscence, line 10: same for this

src/services/cherry-picker.ts
```diff
@@ -218,6 +235,57 @@ export class CherryPicker {
     await this.redis.set(blockchain + '-' + id + '-service', JSON.stringify(serviceQuality), 'EX', ttl)
   }
 
+  /**
+   * Nodes may fail ocasionally due to misconfigured timeouts, this can specially
+   * hurt on archival chains where big operations require more time (and grater
```
nymd: * greater

src/services/cherry-picker.ts
```diff
@@ -218,6 +235,57 @@ export class CherryPicker {
...
+   * timeouts) to operate. When a node continously fail due to the misconfiguration,
+   * is removed from the session
+   * @param blockchain
+   * @param serviceNode
+   * @param elapsedTime
+   * @param timeout
```
nymd: I think this should be timeoutCounter or something similar.

nymd: or even just `timeouts`

nymd (changes requested):

src/application.ts
```diff
@@ -77,7 +78,8 @@ export class PocketGatewayApplication extends BootMixin(ServiceMixin(RepositoryM
     const defaultLogLimitBlocks: number = parseInt(DEFAULT_LOG_LIMIT_BLOCKS) || 10000
     const aatPlan = AAT_PLAN || AatPlans.PREMIUM
     const redirects: string | object[] = REDIRECTS || ''
-    const commitHash: string | string = COMMIT_HASH || ''
+    const commitHash: string = COMMIT_HASH || ''
+    const archivalChains: string[] = (ARCHIVAL_CHAINS || []).split(',')
```
nymd: Will this fail if someone accidentally puts a space after the commas?

nymd (approved):

src/application.ts
rem1niscence: not fail per se but will add the space to the result, will add a small fix for that

PR #340, opened 2021-10-25
change redis timeout to remaining of hour

rem1niscence:
These are the changes, bear in mind this branch is based on the commit right before the pocket-relayer refactor:
1. bug: exhausted nodes were only filtered if there were 2 or more exhausted, not at least one
2. bug: The ttl was updated to two hours from now  every time there was not at least two nodes exhausted
3. bug: Also, the dummy value was also added every time innecesarily
4. feat: all the needed values are added at most once
5. feat: the ttl is set to the end of the hour + 2 minutes threshold. 

Also, this changes a lot of signatures of some common functions (particurlarly `recordMetric`, `consensusFilter` and `chainIDFilter` along with internal functions of the classes that exposes them) this is due to the new sessionKey format which dumps the sessionKey provided by the session and instead hashes the nodes public keys like it used to before, files and tests have been changed accordingly


src/services/cherry-picker.ts
```diff
@@ -172,10 +172,10 @@ export class CherryPicker {
     elapsedTime: number,
     result: number,
     timeout?: number,
-    sessionKey?: string
+    pocketSession?: Session
   ): Promise<void> {
-    await this._updateServiceQuality(blockchain, applicationID, elapsedTime, result, 900, timeout, sessionKey)
-    await this._updateServiceQuality(blockchain, serviceNode, elapsedTime, result, 7200, timeout, sessionKey)
+    await this._updateServiceQuality(blockchain, applicationID, elapsedTime, result, 900, timeout, pocketSession)
```
nymd: It looks like here we are still using "blockchain" when really what we mean is "blockchainID" -- is that right?  Can we get these variables to match all the way through, thanks

src/services/cherry-picker.ts
```diff
@@ -217,15 +217,15 @@ export class CherryPicker {
         ) // divided by total results
           .toFixed(5) // to 5 decimal points
       } else {
-        await this.updateBadNodeTimeoutQuality(blockchain, id, elapsedTime, timeout, sessionKey)
+        await this.updateBadNodeTimeoutQuality(blockchain, id, elapsedTime, timeout, pocketSession)
```
nymd: Same here, blockchain vs blockchainID

src/services/metrics-recorder.ts
```diff
@@ -156,7 +159,7 @@ export class MetricsRecorder {
           elapsedTime,
           result,
           timeout,
-          sessionKey
+          pocketSession
```
nymd, line 164: This needs to be a string -- the hashed sessionKey -- here it is the full session going to the logs.

src/utils/cache.ts
```diff
@@ -11,15 +14,30 @@ const ALTRUIST_URL = JSON.parse(process.env.ALTRUISTS)?.['0001']
...
-  await redis.sadd(`session-${sessionKey}`, nodePubKey)
-  await redis.del(`sync-check-${sessionKey}`, `chain-check-${sessionKey}`)
-}
+export async function removeNodeFromSession(
+  redis: Redis,
+  blockchain: string,
+  sessionNodes: Node[],
+  nodePubKey: string
+): Promise<void> {
+  const hash = hashBlockchainNodes(blockchain, sessionNodes)
```
nymd: blockchainID

src/utils/helpers.ts
```diff
@@ -0,0 +1,15 @@
+import { Node } from '@pokt-network/pocket-js'
+const crypto = require('crypto')
+
+// hashes a blockchain and all of the nodes given, sorted by public key
+export function hashBlockchainNodes(blockchain: string, nodes: Node[]): string {
+  return `${blockchain}-${crypto
```
nymd: or is it ID?

nymd (changes requested):

src/services/metrics-recorder.ts
rem1niscence: on this case inside the function the variable is destructure and uses the `sessionKey` and `sessionNodes` whenever is necessary, otherwise would have to pass both variables which would achieve the same thing

nymd: But this is the old pocket session key being sent there, the one that never changes:

https://github.com/pokt-foundation/portal-api/blob/523e85133a15a51a9a507b9899549270974db638/src/services/metrics-recorder.ts#L86

nymd: The logger should be using the hashed version, not the one destructured from PocketSession that never changes, right?

nymd (changes requested):

nymd (approved):

PR #353, opened 2021-11-05
back to a working session key, never trust pocket js

nymd:


src/services/pocket-relayer.ts
```diff
@@ -442,20 +443,24 @@ export class PocketRelayer {
...
+      // sessionKey = "blockchain and a hash of the all the nodes in this session, sorted by public key"
+      const sessionKey = `${blockchain}-${crypto
+        .createHash('sha256')
+        .update(
+          JSON.stringify(
+            nodes.sort((a, b) => (a.publicKey > b.publicKey ? 1 : b.publicKey > a.publicKey ? -1 : 0)),
+            (k, v) => (k !== 'publicKey' ? v : undefined)
+          )
+        )
+        .digest('hex')}`
```
Evalir, line 460: This returns a string, right? Just to make extra extra sure

Evalir (dismissed):
so sleek, so modern

src/services/pocket-relayer.ts
nymd: It did for the first 18 months it was in use in the codebase

crisog (approved):

Evalir (approved):
k now this looks good!

PR #370, opened 2021-11-10
Automatic require envs

rem1niscence:
Small utility PR which removes the manual check whether required envs are added and offloads to a single place, just need to add the needed env in one of the three lists `requiredEnvVars`, `requiredEnvVarsOnlyInProd` and `optionalEnvVars` and you're good to go


nymd (approved):

src/application.ts
```diff
@@ -69,9 +69,6 @@ export class PocketGatewayApplication extends BootMixin(ServiceMixin(RepositoryM
       INFLUX_URL,
       INFLUX_TOKEN,
       INFLUX_ORG,
-      AWS_ACCESS_KEY_ID,
```
nymd, line 72: Wait, why did we lose these?  they are now required for timestream

nymd (changes requested):

src/application.ts
rem1niscence: they're now only checked on `environment.observer.ts` they will still throw error if not added, see here: https://github.com/pokt-foundation/portal-api/blob/a60a159f1359b1a31fc9116f4a1094db0ecb7b07/src/observers/environment.observer.ts#L41

PR #381, opened 2021-11-11
Always relay to altruist when node relay fails

rem1niscence:
Adds a general try-catch to the node sendRelay process, which allows for the process to continue in case of any failure, also changes the `getCurrentSession` have a maximum number of retries with a different timeout than default.

Note: this PR is currently using the staging version of pocket-js which has not been release yet on npm

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,7 +71,7 @@
     "@loopback/rest": "^9.3.1",
     "@loopback/rest-explorer": "3.3.1",
     "@loopback/service-proxy": "^3.2.1",
-    "@pokt-network/pocket-js": "0.6.16-rc",
+    "@pokt-network/pocket-js": "https://github.com/pokt-network/pocket-js.git",
     "aws-sdk": "^2.1023.0",
     "axios": "^0.21.1",
     "dotenv": "^8.2.0",
--- a/src/services/pocket-relayer.ts
+++ b/src/services/pocket-relayer.ts
@@ -176,125 +176,135 @@ export class PocketRelayer {
     const method = parseMethod(parsedRawData)
     const fallbackAvailable = this.altruists[blockchainID] !== undefined ? true : false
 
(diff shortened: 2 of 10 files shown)


nymd (dismissed):

package.json
```diff
@@ -71,7 +71,7 @@
     "@loopback/rest": "^9.3.1",
     "@loopback/rest-explorer": "3.3.1",
     "@loopback/service-proxy": "^3.2.1",
-    "@pokt-network/pocket-js": "0.6.16-rc",
+    "@pokt-network/pocket-js": "https://github.com/pokt-network/pocket-js.git",
```
crisog, line 74:
```suggestion
    "@pokt-network/pocket-js": "https://github.com/pokt-network/pocket-js.git#staging",
```

nymd (approved):

src/services/pocket-relayer.ts
```diff
@@ -479,261 +494,269 @@ export class PocketRelayer {
...
+    const nodesToRemove = await this.redis.smembers(sessionCacheKey)
+
+    if (nodesToRemove.length > 0) {
+      nodes = nodes.filter((n) => !nodesToRemove.includes(n.publicKey))
+    }
 
-      let syncCheckPromise: Promise<Node[]>
-      let syncCheckedNodes: Node[]
+    if (nodes.length === 0) {
+      logger.log('warn', `SESSION: blockchainHash: ${sessionKey} has exhausted all node relays`, {
```
crisog: I'd prefer leaving it as before, blockchainHash in the string doesn't provide much value in that log.

`SESSION: ${sessionKey} blah blah`

crisog (approved):

package.json
nymd: Staging is default and I won't be messing with that but yes that is a good point.

rem1niscence: can be left as is right now yes

PR #382, opened 2021-11-12
feat: increase granularity of timestream log

rem1niscence:
hard PR, check carefully

### DIFF
--- a/src/services/metrics-recorder.ts
+++ b/src/services/metrics-recorder.ts
@@ -213,7 +213,7 @@ export class MetricsRecorder {
       const timeStreamMeasure = {
         Dimensions: timeStreamDimensions,
         MeasureName: 'elapsedTime',
-        MeasureValue: `${parseFloat(elapsedTime.toFixed(4))}`,
+        MeasureValue: `${parseFloat(elapsedTime.toString())}`,
         MeasureValueType: 'DOUBLE',
         Time: Date.now().toString(),
       }


src/services/metrics-recorder.ts
```diff
@@ -213,7 +213,7 @@ export class MetricsRecorder {
       const timeStreamMeasure = {
         Dimensions: timeStreamDimensions,
         MeasureName: 'elapsedTime',
-        MeasureValue: `${parseFloat(elapsedTime.toFixed(4))}`,
+        MeasureValue: `${parseFloat(elapsedTime.toString())}`,
```
nymd, line 216: wrong time :D  granularity of the timestamp, not this one

nymd (changes requested):

src/services/metrics-recorder.ts
nymd: image

nymd: That time on the right col

rem1niscence: on that case both can be changed, according to the error:

```
  {
    Reason: 'A record already exists with the same time, dimensions, measure name, and record version. A higher record version must be specified in order to update the measure value. Specifying record version is supported by the latest SDKs.',
    RecordIndex: 0
  }
```
dimension values are also taken into account, but that's good so even more granularity can be achieved, will do the change

rem1niscence: After a bit of research came to the conclusion that nodejs doesn't natively supports microseconds and after trying a few recommendations they didn't work reliably, recommend at the moment merge with only this change to provide slightly better granularity

nymd (approved):

PR #390, opened 2021-11-13
Checks rewrite

rem1niscence:
Provides a complete rewrite of the checks services (chain, sync and now archival) with the main feat being that the base service to perform the check does not need any external dependencies,  the external dependencies are introduced in wrapper classes over the checks while reusing several functions with a bit of OOP to reduce duplication. Also, this  PR also has the archival check added but is not currently applied, will leave it for another PR after this is battled tested.

The main motivation behind this PR is that currently, the way the checks are handled produces a great amount of code duplication for behavior that is completely the same, also, all the project-specific dependencies are bundled together as part of the checks code, which makes it really hard to use the code in other external  environments. With this, all the divisions are specifically wrapped over and provides a solid base so the checks core code can be later exported as an external package. Leaving a clear division in what is needed to perform a check and what is specific of the portal-api. 

A few things to consider:
 * Right now both the current checks and the new ones are in the code, the new checks are suffixed with the word `new` on their files and have a prefix `Pocket` in their class, that and other some minor changes are planned to be removed once the old checks can be safely deprecated
 * This maintains <del>99%</del> 100% of the current functionality, <del>only issue right now is when logging the elapse time of the nodes to perform a check, logs are the same for all of them (as in the one that took the longest), that's an issue I haven't found a solution yet, everything else is exactly as the current checks</del>. Fixed :)
 * Also have some small linting fixes as eslint randomly started whining about everything 
 * When possible, tests haven not been changed but only added the new signature of the checks, everything else left untouched to reassure same behavior.

I expect a lot of comments to come from this PR so please check it carefully.


src/controllers/v1.controller.ts
```diff
@@ -71,8 +71,15 @@ export class V1Controller {
...
     })
-    this.syncChecker = new SyncChecker(this.redis, this.metricsRecorder, this.defaultSyncAllowance, this.origin)
-    this.chainChecker = new ChainChecker(this.redis, this.metricsRecorder, this.origin)
+    this.syncChecker = new PocketSyncChecker(
+      this.pocket,
+      this.redis,
+      this.metricsRecorder,
+      this.origin,
+      this.defaultSyncAllowance
+    )
```
Evalir: This is a nitpick, but I'd prefer having an object here—this is prone to mistakes due to ordering being important, even though we have typescript.

src/services/archival-check.ts
```diff
@@ -0,0 +1,113 @@
+import { Redis } from 'ioredis'
+import { Pocket, PocketAAT, Configuration, Node, Session } from '@pokt-network/pocket-js'
```
Evalir: Let's remember to organize alphabetically these imports in parenthesis. :)

src/services/archival-check.ts
```diff
@@ -0,0 +1,113 @@
...
+    nodes: Node[],
+    archivalCheckOptions: ArchivalCheckOptions,
+    blockchainID: string,
+    pocketAAT: PocketAAT,
+    pocketConfiguration: Configuration | undefined,
+    pocketSession: Session,
+    applicationID: string,
+    applicationPublicKey: string,
+    requestID: string
+  ): Promise<Node[]> {
```
Evalir: I'd also make this an object to avoid mistakes due to ordering.

src/services/chain-checker-new.ts
```diff
@@ -0,0 +1,106 @@
+import { Redis } from 'ioredis'
+import { Configuration, Node, Pocket, PocketAAT, Session } from '@pokt-network/pocket-js'
```
Evalir, line 2: :)

src/services/node-checker-wrapper.ts
```diff
@@ -0,0 +1,272 @@
+import { Redis } from 'ioredis'
+import { Pocket, Node, PocketAAT, Configuration, Session } from '@pokt-network/pocket-js'
```
Evalir, line 2: :)

src/services/node-checker-wrapper.ts
```diff
@@ -0,0 +1,272 @@
...
+  }
+
+  /**
+   * Helper method to check for cached checks, is a check is already cached or in progress,
+   * returns valid nodes from the cached result or session. Otherwise set a cache lock.
+   * @param {Node[]} nodes session nodes.
+   * @param cacheKey key to get/set results.
+   * @returns nodes cached or provided on case of cache/lock.
+   */
+  protected async cacheNodes(nodes: Node[], cacheKey: string): Promise<Node[]> {
```
Evalir: This function doesn't cache nodes; it checks if they're cached and retrieves them or sets a cache lock. I think this function name could be a little bit better!

src/services/node-checker-wrapper.ts
```diff
@@ -0,0 +1,272 @@
...
+    checkType: Check,
+    nodes: Node[],
+    nodesPromise: PromiseSettledResult<NodeCheckResponse<T>>[],
+    blockchainID: string,
+    pocketSession: Session,
+    requestID: string,
+    relayStart: [number, number],
+    applicationID: string,
+    applicationPublicKey: string
+  ): Promise<NodeCheckResponse<T>[]> {
```
Evalir: Again, I'd also make this an object.

src/services/node-checker-wrapper.ts
```diff
@@ -0,0 +1,272 @@
...
+   * @param requestID request id.
+   * @param relayStart time when the checks started.
+   * @param applicationID application database's ID.
+   * @param applicationPublicKey application's public key.
+   * @returns nodes having successful and valid check results.
+   */
+  protected async filterNodes<T>(
+    checkType: Check,
+    nodes: Node[],
+    nodesPromise: PromiseSettledResult<NodeCheckResponse<T>>[],
```
Evalir: This variable name could be a little bit more descriptive—the "type" of it doesn't need to be in the name, as it can be inferred from its typescript type.

src/services/node-checker-wrapper.ts
```diff
@@ -0,0 +1,272 @@
...
+    const { sessionNodes } = pocketSession
+    const sessionHash = hashBlockchainNodes(blockchainID, sessionNodes)
+
+    for (const [idx, nodeCheckPromise] of nodesPromise.entries()) {
+      const node = nodes[idx]
+
+      const { serviceURL, serviceDomain } = await getNodeNetworkData(this.redis, node.publicKey, requestID)
+
+      // helps debugging
+      const formattedType = checkType.replace('-', ' ').toUpperCase()
```
Evalir, line 96: Wonder if instead of doing this we should just have an object (or a set of `Symbol`s) with the different check types? That way it's also impossible to fat-finger them.

src/services/node-checker-wrapper.ts
```diff
@@ -0,0 +1,272 @@
...
+
+      const rejected = nodeCheckPromise.status === 'rejected'
+      const failed = rejected || nodeCheckPromise.value.response instanceof Error
+
+      // Error
```
Evalir: Don't think we need this as the variable name indicates it!

src/services/node-checker-wrapper.ts
```diff
@@ -0,0 +1,272 @@
...
+          result: 500,
+          delivered: false,
+          fallback: false,
+          method: checkType,
+          error: (error as string) || errorMsg,
+          origin: this.origin,
+          data: undefined,
+          sessionHash,
+          bytes: 0,
+          pocketSession: pocketSession,
```
Evalir: We could probably order this alphabetically as well :).

src/services/node-checker.ts
```diff
@@ -0,0 +1,376 @@
+import get from 'lodash/get'
+import { Pocket, Configuration, Node, RelayResponse, PocketAAT, HTTPMethod, RpcError } from '@pokt-network/pocket-js'
```
Evalir, line 2: :)

src/services/node-checker.ts
```diff
@@ -0,0 +1,376 @@
...
+   * @param node Node to check.
+   * @param data Payload to send to the blockchain, expected to return the blockchain's chainID.
+   * @param blockchainID Blockchain to request data from.
+   * @param aat Pocket Authentication Token object.
+   * @param chainID  blockchain chain's ID to evaluate against.
+   * @param path  optional. Blockchain's path to send the request to.
+   * @returns Response object containing the relay response, request output and boolean
+   * assuring whether the node supports the correct chain or not
+   */
+  async chain(
```
Evalir: Shouldn't we name this `checkChain` or `performChainCheck`?

src/services/node-checker.ts
```diff
@@ -0,0 +1,376 @@
...
+    blockchainID: string,
+    aat: PocketAAT,
+    chainID: number,
+    path?: string
+  ): Promise<NodeCheckResponse<ChainCheck>> {
+    let nodeChainID: number
+
+    const isCorrectChain = (payload: object, chainIDArg) => {
+      nodeChainID = NodeChecker.parseBlockFromPayload(payload, 'result')
+      return nodeChainID === chainIDArg
```
Evalir: Any reason this function needs to be an arrow function, and be inside another function?

src/services/node-checker.ts
```diff
@@ -0,0 +1,376 @@
...
+    blockchainID: string,
+    aat: PocketAAT,
+    resultKey: string,
+    path?: string,
+    source?: number,
+    allowance = 0
+  ): Promise<NodeCheckResponse<SyncCheck>> {
+    let blockheight: number
+
+    const isSynced = (payload: object, minimumAllowedHeight) => {
```
Evalir: Same with this function

src/services/node-checker.ts
```diff
@@ -0,0 +1,376 @@
...
+    blockchainID: string,
+    aat: PocketAAT,
+    resultKey: string,
+    comparator: string | number,
+    path?: string,
+    swap = false
+  ): Promise<NodeCheckResponse<ArchivalCheck>> {
+    let payloadResponse: object
+
+    const isArchival = (payload: object, comparatorVal: string) => {
```
Evalir: And this function

src/services/node-checker.ts
```diff
@@ -0,0 +1,376 @@
...
+      success,
+      response: relayResponse.payload,
+      output: {
+        message: JSON.stringify(payloadResponse),
+      },
+    }
+  }
+
+  /**
+   * Helper function for request the blockchains data, asserting is valid and return the result from a comparator function
```
Evalir:
```suggestion
   * Helper function for requesting the blockchain data, asserting it's valid, and returning the result from a comparator function
```

src/services/node-checker.ts
```diff
@@ -0,0 +1,376 @@
...
+   * @param pocketConfiguration Pocket's Configuration object.
+   * @returns updated configuration object.
+   */
+  private updateConfigurationConsensus(pocketConfiguration: Configuration): Configuration {
+    return new Configuration(
+      pocketConfiguration.maxDispatchers,
+      pocketConfiguration.maxSessions,
+      5,
+      2000,
+      false,
```
Evalir: I think we could lift these arbitrary values into constants at the top of the file.

src/services/sync-checker-new.ts
```diff
@@ -0,0 +1,318 @@
...
+        'SYNC CHECK CHALLENGE:',
+        requestID
+      )
+    }
+
+    return syncedNodes
+  }
+
+  /**
+   * Obtains the blockheight from ann altruist node.
```
Evalir:
```suggestion
   * Obtains the blockheight from an altruist node.
```

tests/unit/archival-check.unit.ts
```diff
@@ -0,0 +1,176 @@
+import axios from 'axios'
+import MockAdapter from 'axios-mock-adapter'
+import RedisMock from 'ioredis-mock'
+import { expect, sinon } from '@loopback/testlab'
+import { Configuration, Session, RpcError } from '@pokt-network/pocket-js'
+
```
Evalir: Let's either remove this line or add it everywhere else, to keep consistency on imports :).

Evalir (commented):
Haven't put too too much effort yet into scrutinising the overall architecture but I'm liking what I'm seeing so far. Tomorrow I'll have another, more in-depth pass; left some nitpicks in the meanwhile.

Evalir:
Also, I know this is a big ask, but do you think you could add the motivation behind this PR, and why this architecture is better? Even if this becomes obvious after reviewing, I think it can help speed the review time up.

src/services/node-checker.ts
```diff
@@ -0,0 +1,376 @@
...
+  }
+
+  /**
+   * Helper function for request the blockchains data, asserting is valid and return the result from a comparator function
+   * over the obtained relay response.
+   * @param node node to perfom the request.
+   * @param data payload to send to the blockchain.
+   * @param blockchainID Blockchain to request data from.
+   * @param aat Pocket Authentication token object.
+   * @param resultKey key to extract data from the JSON response, for nested keys can be added using dot notation
```
crisog: This should be `path` instead of `resultKey`

src/services/node-checker-wrapper.ts
```diff
@@ -0,0 +1,272 @@
...
+    }
+
+    return checkedNodes
+  }
+
+  /**
+   * Logs responses from the node checks and filters failing nodes.
+   * @param checkType type of check made.
+   * @param nodes nodes to be filtered.
+   * @param nodesPromise results of a check on node's.
```
crisog:
```suggestion
   * @param nodesPromise results of a check on nodes.
```

src/services/sync-checker-new.ts
```diff
@@ -0,0 +1,318 @@
...
+    blockchainID: string,
+    pocketAAT: PocketAAT,
+    pocketConfiguration: Configuration | undefined,
+    pocketSession: Session,
+    blockchainSyncBackup: string,
+    applicationID: string,
+    applicationPublicKey: string,
+    requestID: string
+  ): Promise<Node[]> {
+    const sessionHash = hashBlockchainNodes(blockchainID, pocketSession.sessionNodes)
```
crisog, line 49: I'm pretty sure this same thing is also called `sessionKey` in other places; if we are going to replace it with `sessionHash` let's make sure it's consistent across the codebase.

src/services/node-checker.ts
crisog: +1 to `performChainCheck`.

src/services/sync-checker-new.ts
```diff
@@ -0,0 +1,318 @@
...
+            requestID: requestID,
+            serviceNode: publicKey,
+            blockchainID,
+            origin: this.origin,
+            serviceURL,
+            serviceDomain,
+            sessionHash,
+          })
+
+          this.metricsRecorder
```
crisog, line 213: Isn't this already being handled by `filterNodes()` at `NodeCheckerWrapper`?

src/services/archival-check.ts
```diff
@@ -0,0 +1,113 @@
...
+
+  /**
+   * Perfoms an archival check on all the nodes provided, this is nodes must confirm they're able to relay
+   * archival request, failing to do so will result in the node getting slashed.
+   * @param nodes nodes to perfom the check on.
+   * @param archivalCheckOptions. options containing the blockchain's archival configuration.
+   * @param blockchainID Blockchain to request data from.
+   * @param pocketAAT Pocket Authentication Token object.
+   * @param pocketConfiguration pocket's configuration object.
+   * @param pocketSession. pocket's current session object.
```
crisog:
```suggestion
   * @param pocketSession pocket's current session object.
```

src/services/archival-check.ts
```diff
@@ -0,0 +1,113 @@
...
+   * @param applicationID application database's ID.
```
crisog:
```suggestion
   * @param applicationID application's database ID.
```

src/services/archival-check.ts
```diff
@@ -0,0 +1,113 @@
...
+export class ArchivalChecker extends NodeCheckerWrapper {
+  constructor(pocket: Pocket, redis: Redis, metricsRecorder: MetricsRecorder, origin: string) {
+    super(pocket, redis, metricsRecorder, origin)
+  }
+
+  /**
+   * Perfoms an archival check on all the nodes provided, this is nodes must confirm they're able to relay
+   * archival request, failing to do so will result in the node getting slashed.
+   * @param nodes nodes to perfom the check on.
+   * @param archivalCheckOptions. options containing the blockchain's archival configuration.
```
crisog:
```suggestion
   * @param archivalCheckOptions options containing the blockchain's archival configuration.
```

src/services/node-checker.ts
```diff
@@ -0,0 +1,376 @@
...
+    resultKey: string,
+    comparator: string | number,
+    path?: string,
+    swap = false
+  ): Promise<NodeCheckResponse<ArchivalCheck>> {
+    let payloadResponse: object
+
+    const isArchival = (payload: object, comparatorVal: string) => {
+      payloadResponse = payload
+      const result = NodeChecker.parseBlockFromPayload(payload, resultKey).toString()
```
crisog: If we are using `eth_getBalance` call for the archival check, then we receive balance as a hex number - OK. I think it's time to rename `parseBlockFromPayload`. 

`parseHexFromPayload` sounds like a better name better name, generalizing it so it can be used for both blocks & balances without causing confusion; at the end, it's just hex numbers.

crisog:
None of my previous review comments indicate errors/failure in the logic; the PR looks good overall. Checks are way easier to add now; and the logic for existing checks is pretty much the same (as it should cause they were doing ok).

Thx for the improvement :)

src/services/sync-checker-new.ts
rem1niscence: `sessionKey` at this moment is only used in pocket-relayer to refer as hash, so you're right, that only changed can be done in other PR as is not related to this

src/services/node-checker-wrapper.ts
rem1niscence: On that case it could be just `filter` then as is what it does besides logging

crisog: > On that case it could be just `filter` then as is what it does besides logging

I think he meant the `nodesPromise` variable.

rem1niscence: you're right lol

src/services/node-checker-wrapper.ts
rem1niscence: they're part of a type (as an enum) [from node-checker](https://github.com/pokt-foundation/portal-api/blob/54afc9fd085f04337e7f5fdf4ce9c2b78ad35c1f/src/services/node-checker.ts#L6) writing improperly will trigger a compiler error and autocomplete on an editor with intellisense-like features

src/services/node-checker.ts
rem1niscence:
```
const nodeChecker = new nodeChecker(params...)
nodeChecker.chain() 
```

both names seem redundant to me

src/services/node-checker.ts
rem1niscence: I use arrow functions unless I explicitly need the previous `this` context, helps from my pov reduce unexpected side effects, also is inside the function as a first-class object (variable on this case) to be passed as a param to `processCheck`

src/services/node-checker.ts
rem1niscence: check answer above

src/services/node-checker.ts
rem1niscence: check answer above

src/services/node-checker.ts
rem1niscence: Oxford commas  ඞ

src/services/node-checker.ts
Evalir: I don't think its redudant at all—and even if it might seem a little bit redundant, having no context is even worse. `nodeChecker.chain()` doesn't tell me much about what this is doing, meanwhile `performChainCheck` does.

src/services/node-checker.ts
Evalir: The thing is, the fact that you're using an arrow function instead a function declaration means that you're _adding_, not removing, unexpected side-effects, as arrow functions arrived later than function declarations :). As we don't need the `this`, it's a matter of style, but I personally dislike defining utility functions like these inside a method, when they perfectly could go in a `checker-utilts.ts` file or just defined at the top of the file.

src/services/sync-checker-new.ts
rem1niscence: nope, the sync check logs on `filterNode` are excluded from sync-check as it has more validations to perform

src/services/node-checker-wrapper.ts
Evalir: Yup, if it's a type sg as well

Evalir:
Will re-review today so we can get going with this PR again—even though we might want to delay this being introduced today due to the stickiness fix.

rem1niscence:
as the variables like `nodeSyncCheckData` works, will sent the `promiseRejection` value as a payload response, this will caused the node to be logged as error and filtered out

jessicadaugherty:
> Also, this PR also has the archival check added but is not currently applied, will leave it for another PR after this is battled tested.

resurrecting this thread in light of issues with eth_getlogs. as we are revisiting QoS, archival check should be included imo, whether in this pr or a separate QoS related one as part of the enhancement effort. not sure where it fits in with the current conversation @crisog but just flagging it (although probably already on your radar) for team discussion.

PR #411, opened 2021-11-17
dd region tag

rem1niscence:


Evalir (approved):

src/services/logger.ts
```diff
@@ -108,18 +108,21 @@ const options = {
...
+    ddtags: `region:${region}`,
   },
 }
 
 const getTransports = () => {
   const transports = [new winstonTransports.Console(options.console)]
 
   if (environment === 'production' || environment === 'staging') {
-    if (logToCloudWatch) {
+    if (logToCloudWatch || logToDataDog) {
```
nymd, line 119: Did I miss where this was added to .envs and .env example?

const logToDataDog = process.env.LOG_TO_DATADOG === 'true'

nymd (changes requested):

src/services/logger.ts
rem1niscence: was added to envs on a previous PR but not to example, will do the example one

PR #412, opened 2021-11-17
initial try

nymd:


src/controllers/v1.controller.ts
```diff
@@ -284,6 +292,42 @@ export class V1Controller {
...
   }
 
+  async checkClientStickiness(
+    rawData: object
+  ): Promise<{ preferredApplicationID: string; preferredNodeAddress: string; rpcID: number }> {
+    // Parse the raw data to determine the lowest RPC ID in the call
+    const parsedRawData = Object.keys(rawData).length > 0 ? JSON.parse(rawData.toString()) : JSON.stringify(rawData)
+    const rpcID = parseRPCID(parsedRawData)
+
+    if (rpcID > 0) {
```
Evalir, line 302: Bit of a nitpick, but I feel this would be a bit easier to read if we invert this if—If it's 0 or below, just return the default response in line `328`

src/services/pocket-relayer.ts
```diff
@@ -680,7 +692,17 @@ export class PocketRelayer {
...
     }
 
-    const node = await this.cherryPicker.cherryPickNode(application, nodes, blockchainID, requestID)
+    let node: Node
+
+    // Before cherry picking, check to see if preferred node is in the set of good nodes
+    const preferredNodeIndex = nodes.findIndex((x) => x.address === preferredNodeAddress)
+
+    if (preferredNodeAddress && preferredNodeIndex > 0) {
+      console.log('STICKINESS SUCCESS', preferredNodeAddress)
```
rem1niscence, line 701: if these console.logs are meant to be kept then should be changed to `logger.log`, otherwise removed

nymd: Ah yea I did want to see them on staging/canary at first, I will change them.

rem1niscence: as the way it works now, you won't be able to see them on DD unless you do a logger.log, but will on cw

src/controllers/v1.controller.ts
nymd: Disagree -- I also need the failure case at the bottom in case there is an issue inside of the if {}. For instance if the redis key is not found, it needs a failure path.

src/services/pocket-relayer.ts
```diff
@@ -443,12 +453,14 @@ export class PocketRelayer {
     blockchainChainID,
   }: {
     data: string
+    rpcID: number | 0
```
rem1niscence, line 456: no big deal but this is redundant, can be just `number`

PR #430, opened 2021-11-18
Feature/sc 63/polish client node stickiness

rem1niscence:
* Adds lb/app config for stickiness
* Modified pocket mock to return node public key
* Remove sticky logs in favor of single boolean on metrics recorder which is `true` if the call was a sticky success
* Add unit and acceptance test

### DIFF
--- a/src/controllers/v1.controller.ts
+++ b/src/controllers/v1.controller.ts
@@ -178,14 +178,13 @@ export class V1Controller {
 
       const { stickiness = false, stickinessDuration = DEFAULT_STICKINESS_DURATION } = loadBalancer
 
-      const stickinessData = stickiness
+      const { preferredApplicationID, preferredNodeAddress, rpcID } = stickiness
         ? await this.checkClientStickiness(rawData)
         : {
             preferredApplicationID: '',
             preferredNodeAddress: '',
             rpcID: 0,
           }
-      const { preferredApplicationID, preferredNodeAddress, rpcID } = stickinessData
 
       const application = await this.fetchLoadBalancerApplication(
         loadBalancer.id,
@@ -275,13 +274,9 @@ export class V1Controller {
       if (application?.id) {
         const { stickiness = false, stickinessDuration = DEFAULT_STICKINESS_DURATION } = application
 
-        const stickinessData = stickiness
+        const { preferredNodeAddress, rpcID } = stickiness
           ? await this.checkClientStickiness(rawData)
-          : {
-              preferredNodeAddress: '',
-              rpcID: 0,
-            }
-        const { preferredNodeAddress, rpcID } = stickinessData
+          : { preferredNodeAddress: '', rpcID: 0 }
 
         const sendRelayOptions: SendRelayOptions = {
           rawData,
--- a/src/controllers/v1.controller.ts
+++ b/src/controllers/v1.controller.ts
@@ -19,7 +19,12 @@ import { loadBlockchain } from '../utils/relayer'
 import { SendRelayOptions } from '../utils/types'
 const logger = require('../services/logger')
 
-const DEFAULT_STICKINESS_DURATION = 300
+const DEFAULT_STICKINESS_DURATION = 300 // Seconds
+const DEFAULT_STICKINESS_PARAMS = {
+  preferredApplicationID: '',
+  preferredNodeAddress: '',
+  rpcID: 0,
+}
 
 export class V1Controller {
   cherryPicker: CherryPicker
@@ -180,11 +185,7 @@ export class V1Controller {
 
       const { preferredApplicationID, preferredNodeAddress, rpcID } = stickiness
         ? await this.checkClientStickiness(rawData)
-        : {
-            preferredApplicationID: '',
-            preferredNodeAddress: '',
-            rpcID: 0,
-          }
+        : DEFAULT_STICKINESS_PARAMS
 
       const application = await this.fetchLoadBalancerApplication(
         loadBalancer.id,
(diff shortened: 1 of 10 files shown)


src/controllers/v1.controller.ts
```diff
@@ -173,7 +175,16 @@ export class V1Controller {
 
       // Fetch applications contained in this Load Balancer. Verify they exist and choose
       // one randomly for the relay. First check RPC ID to see if this client should be stuck to an app and node.
-      const { preferredApplicationID, preferredNodeAddress, rpcID } = await this.checkClientStickiness(rawData)
+
+      const { stickiness = false, stickinessDuration = DEFAULT_STICKINESS_DURATION } = loadBalancer
+
+      const { preferredApplicationID, preferredNodeAddress, rpcID } = stickiness
```
crisog, line 181: Looks a bit confusing; stickiness is a boolean. How are we getting these values from it?

src/controllers/v1.controller.ts
```diff
@@ -173,7 +175,16 @@ export class V1Controller {
...
+        ? await this.checkClientStickiness(rawData)
+        : {
+            preferredApplicationID: '',
+            preferredNodeAddress: '',
+            rpcID: 0,
+          }
```
crisog: I'd prefer it this way; but feel free to keep it like it is now as the logic is the same.
```suggestion
      const stickinessData = stickiness
        ? await this.checkClientStickiness(rawData)
        : {
            preferredApplicationID: '',
            preferredNodeAddress: '',
            rpcID: 0,
          }

      const { preferredApplicationID, preferredNodeAddress, rpcID } = stickinessData
```

src/controllers/v1.controller.ts
```diff
@@ -258,14 +270,20 @@ export class V1Controller {
 
     try {
       const application = await this.fetchApplication(id, filter)
-      const { preferredNodeAddress, rpcID } = await this.checkClientStickiness(rawData)
 
       if (application?.id) {
+        const { stickiness = false, stickinessDuration = DEFAULT_STICKINESS_DURATION } = application
+
+        const { preferredNodeAddress, rpcID } = stickiness
```
crisog, line 278: Same as above

src/controllers/v1.controller.ts
nymd: For me the original way reads fine and is possibly clearer.

src/controllers/v1.controller.ts
```diff
@@ -19,6 +19,8 @@ import { loadBlockchain } from '../utils/relayer'
 import { SendRelayOptions } from '../utils/types'
 const logger = require('../services/logger')
 
+const DEFAULT_STICKINESS_DURATION = 300
```
Evalir: Is this in seconds?

src/controllers/v1.controller.ts
```diff
@@ -173,7 +175,17 @@ export class V1Controller {
...
+
+      const { stickiness = false, stickinessDuration = DEFAULT_STICKINESS_DURATION } = loadBalancer
+
+      const stickinessData = stickiness
+        ? await this.checkClientStickiness(rawData)
+        : {
+            preferredApplicationID: '',
+            preferredNodeAddress: '',
+            rpcID: 0,
+          }
```
Evalir: We could probably lift this up to the top of the file as a constant, and name it `DEFAULT_STICKINESS_PARAMS` or something similar.

src/controllers/v1.controller.ts
```diff
@@ -258,14 +271,24 @@ export class V1Controller {
...
 
       if (application?.id) {
+        const { stickiness = false, stickinessDuration = DEFAULT_STICKINESS_DURATION } = application
+
+        const stickinessData = stickiness
+          ? await this.checkClientStickiness(rawData)
+          : {
+              preferredNodeAddress: '',
+              rpcID: 0,
+            }
```
Evalir: same with the comment above (they should be the same object)

src/services/pocket-relayer.ts
```diff
@@ -1,5 +1,6 @@
 import axios, { AxiosRequestConfig, Method } from 'axios'
 import { Redis } from 'ioredis'
+import { getAddressFromPublicKey } from 'pocket-tools'
```
Evalir, line 3: yay :D

src/controllers/v1.controller.ts
nymd: As it is a redis key TTL, yes it is. A comment here behind it `// in seconds` would help.

src/controllers/v1.controller.ts
nymd: If it is repeated code, yes, I agree.

src/services/pocket-relayer.ts
```diff
@@ -238,6 +241,7 @@ export class PocketRelayer {
                 origin: this.origin,
                 data,
                 pocketSession: this.pocketSession,
+                sticky: preferredNodeAddress === (await getAddressFromPublicKey(relayResponse.proof.servicerPubKey)),
```
nymd: `sticky` in the metricsRecorder is defined as a boolean but here you're sending it the address?

I think the better measure would be: `SUCCESS` or `FAILURE` or `NONE`

Success: if stickiness was desired and successful

Failure: if stickiness was desired but unsuccessful

None: not desired

rem1niscence:
Cannot reply the post above as is a comment so will say it here: I'm returning the result of a comparison of the used node between the preferred one, which will return a boolean, but yes I do like returning also none which would be most of the responses to not confuse with false as in stickiness failed

src/services/pocket-relayer.ts
```diff
@@ -833,4 +833,15 @@ export class PocketRelayer {
       pocketConfiguration.rejectSelfSignedCertificates
     )
   }
+
+  static async stickyRelayResult(
+    prefferedNodeAddress: string | undefined,
```
nymd: * preferred not preffered

nymd (changes requested):
SO CLOSE

nymd (approved):

PR #436, opened 2021-11-19
Stickiness: Set rpcID or full sticky

rem1niscence:
Add option to either set stickiness based on rpcID (relays with increment ids will be stickied with the same node) or full sticky (all relays go through the same node), default is rpcID

### DIFF
--- a/src/controllers/v1.controller.ts
+++ b/src/controllers/v1.controller.ts
@@ -14,7 +14,6 @@ import { CherryPicker } from '../services/cherry-picker'
 import { MetricsRecorder } from '../services/metrics-recorder'
 import { PocketRelayer } from '../services/pocket-relayer'
 import { SyncChecker } from '../services/sync-checker'
-import { getNextRPCID } from '../utils/helpers'
 import { parseRPCID } from '../utils/parsing'
 import { loadBlockchain } from '../utils/relayer'
 import { SendRelayOptions } from '../utils/types'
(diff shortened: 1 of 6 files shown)


crisog (dismissed):

src/controllers/v1.controller.ts
```diff
@@ -182,10 +181,11 @@ export class V1Controller {
       // Fetch applications contained in this Load Balancer. Verify they exist and choose
       // one randomly for the relay. First check RPC ID to see if this client should be stuck to an app and node.
 
-      const { stickiness = false, stickinessDuration = DEFAULT_STICKINESS_DURATION } = loadBalancer
+      const { stickiness = false, stickinessDuration = DEFAULT_STICKINESS_DURATION, useRPCID = true } = loadBalancer
+      const stickyKeyPrefix = stickiness && !useRPCID ? loadBalancer?.id : ''
```
nymd, line 188: Please add a comment block here explaining this for future generations

src/services/pocket-relayer.ts
```diff
@@ -848,14 +821,40 @@ export class PocketRelayer {
...
+    applicationID: string,
+    nodeAddress: string,
+    data: string
+  ): Promise<void> {
+    if (!keyPrefix && !rpcID) {
+      return
     }
 
-    return preferredNodeAddress === (await getAddressFromPublicKey(relayNodePublicKey)) ? 'SUCCESS' : 'FAILURE'
+    // If no key prefix is given, set based on rpcID
```
nymd: Better comment here too please -- explain why there might be a key prefix and what it might be

src/services/pocket-relayer.ts
```diff
@@ -848,14 +821,40 @@ export class PocketRelayer {
...
+    if (keyPrefix) {
+      const clientStickyKey = `${keyPrefix}-${this.ipAddress}-${blockchainID}`
+
+      const nextRequest = await this.redis.get(clientStickyKey)
```
nymd, line 846: Another comment here please -- explain that if the key is still set, we don't want to set it again and extend the stickiness

nymd (changes requested):

PR #443, opened 2021-11-22
Sticky remove failure nodes

rem1niscence:
Uses cherry picker's node logs to check for failure, if the stickied node is marked as failure then the stickiness tag is removed from it.

Also moved all the stickiness code to its own util class as was becoming more and more messy.


rem1niscence:
I'm not a huge fan of the class name `NodeSticker` so if you got a better one feel free to recommend

PR #459, opened 2021-11-24
Stickiness origin whitelist

rem1niscence:
Adds optional origin check for stickiness feature

### DIFF
--- a/src/controllers/v1.controller.ts
+++ b/src/controllers/v1.controller.ts
@@ -29,6 +29,7 @@ const DEFAULT_STICKINESS_PARAMS = {
   duration: 30, // seconds
   useRPCID: true,
   relaysLimit: 0,
+  whitelistOrigins: [],
 }
 
 export class V1Controller {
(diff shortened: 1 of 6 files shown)


src/controllers/v1.controller.ts
```diff
@@ -29,6 +29,7 @@ const DEFAULT_STICKINESS_PARAMS = {
   duration: 30, // seconds
   useRPCID: true,
   relaysLimit: 0,
+  whitelistOrigins: [],
```
nymd: Variable name should be changed to reflect stickiness. Right now this sounds like the whitelistOrigins used for security.

stickyOrigins ?

nymd (changes requested):

src/controllers/v1.controller.ts
rem1niscence: uhhmmn is already inside the `stickinessOptions` object, I think that would be redundant

PR #465, opened 2021-11-24
cherry picker refinements

nymd:
Currently the cherry picker is a harsh punishment to nodes that are just slightly slower than the fastest. This causes uneven distribution of relays and node-session exhaustion.

In the current code, the weight factor that you subtract for each subsequent ranked node is hardcoded.

This PR changes that hardcoded value to a dynamic value based on how much slower the node is compared to the previous node.

In the current model, let's assume the nodes have an average elapsed latency of: 

```
node 1  0.25ms
node 2  0.31ms
node 3  0.42ms
node 4  0.45ms
node 5  0.7ms
```

Assuming these nodes all have a similar success/error rate, they will be weighted in buckets like this:

```
node1 10
node2 8
node3 6
node4 4
node5 2
```

Where the fastest node is almost 2x more likely to be selected than the 3rd fastest. This remains constant even if those two speeds are very similar. With the new formula, the distribution is:

```
node1 10
node2 9
node3 8
node4 8
node5 4
```

Which I believe better reflects the distribution of average elapsed times.

### DIFF
--- a/src/services/cherry-picker.ts
+++ b/src/services/cherry-picker.ts
@@ -301,18 +301,18 @@ export class CherryPicker {
     // The app/node with the highest success rate and the lowest average latency will
     // be 10 times more likely to be selected than a node that has had failures.
     let weightFactor = 10
-    let fastestNodeLatency = 0
+    let previousNodeLatency = 0
     let latencyDifference = 0
 
     // This multiplier is tested to produce a curve that adequately punishes slow nodes
-    const weightMultiplier = 8
+    const weightMultiplier = 14
 
     for (const sortedLog of sortedLogs) {
       // Set the benchmark from the fastest node
-      if (!fastestNodeLatency) {
-        fastestNodeLatency = sortedLog.averageSuccessLatency
+      if (!previousNodeLatency) {
+        previousNodeLatency = sortedLog.averageSuccessLatency
       } else {
-        latencyDifference = sortedLog.averageSuccessLatency - fastestNodeLatency
+        latencyDifference = sortedLog.averageSuccessLatency - previousNodeLatency
       }
 
       // The amount you subtract here from the weight factor should be variable based on how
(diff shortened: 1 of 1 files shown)


rem1niscence (approved):

src/services/cherry-picker.ts
```diff
@@ -301,24 +301,47 @@ export class CherryPicker {
...
+        previousNodeLatency = sortedLog.averageSuccessLatency
+      } else {
+        latencyDifference = sortedLog.averageSuccessLatency - previousNodeLatency
+      }
+
+      // The amount you subtract here from the weight factor should be variable based on how
+      // far off this node's average elapsedTime is from the fastest node.
+      // Previously this value was hardcoded 2 in the first bucket
+      if (latencyDifference) {
+        weightFactor = weightFactor - Math.round(latencyDifference * weightMultiplier)
```
Evalir, line 322: Just to understand the quick mafs... how big on average would this `latencyDifference` be? also, I guess that as the logs are "sorted" we don't need to do `Math.abs()` on this right?

I haven't proven it mathematically but I feel we might wanna normalize these numbers somehow? Or am I missing something in the maths?

nymd: Correct they are sorted so no fear of <0. 

Normalize them how?  Need more explanation.

Evalir: > Correct they are sorted so no fear of <0.

sweet! 
> Normalize them how? Need more explanation.

yup scrap that, it mainly applied if things weren't sorted.

on another angle re: huge disparity in latencies:

i'm mainly thinking of edge cases e.g some node being off by 1-2s in latency. How would the formula respond in this case? I'm wondering if there should be some "threshold" were nodes get punished harshly because it might be better for the user to go to fallbacks rather than use the painfully slow nodes. This of course messes with the tokenomics but I'd like to know what you think.

apart from that, this lgtm. If anything I'd leave the explanation of the PR on the code itself (ideally even more detailed) so we can quickly remember how things work 3-4 months of now if we don't touch this code.

nymd: I *think* my comments in the code are pretty good? 

If the latencies are

```
node 1  0.25ms
node 2  0.31ms
node 3  0.42ms
node 4  0.45ms
node 5  1.5ms
```

With node 5 being awful, the distribution would be:

```
node1 10
node2 9
node3 8
node4 8
node5 1
```

nymd: This gives the node a chance to get back in the running without them being selected often.

Evalir: Haha yeah, I see. That's fair:).

Evalir (approved):

crisog (approved):

PR #483, opened 2021-11-26
Cherry picker refinements

nymd:
The cherry picker was written in a simpler time.

Before Sync Check. Before Chain Check. Cherry Picker did all the heavy lifting.

At'll do pig, at'll do. The time for you to try and filter out failing nodes is done. All you really should worry about is latency and recent error rate.

- reduced error log TTLs system wide
- reduced service log TTL from 1 hour to 5 minutes
- stopped counting "Provided Node" error as a penalty against the nodes


src/services/cherry-picker.ts
```diff
@@ -53,7 +53,7 @@ export class CherryPicker {
 
     // Iterate through sorted logs and form in to a weighted list
     // 15 failures per 15 minutes allowed on apps (all 5 nodes failed 3 times)
```
rem1niscence: comment should be changed to 50 instead of 15

src/services/cherry-picker.ts
```diff
@@ -398,7 +397,7 @@ export class CherryPicker {
       errorLog = '0'
     }
 
-    failure = failureLog === 'true' || parseInt(errorLog) > 5
+    failure = failureLog === 'true' || parseInt(errorLog) > 50
```
rem1niscence, line 400: I think this could be better as a constant like `FAILURE_RATE_THRESHOLD`

nymd: Agree, but that holds true for ALL of the things I adjusted with this PR. Not sure I'm ready to commit to making them all variables. Sometimes that makes the code harder to read as you have to reference another file or definitely another spot in the code. For numbers only used once, I think this is tolerable, eh?

rem1niscence (approved):

PR #496, opened 2021-12-01
Use JSON-RPC standard for responses

### LINKED ISSUE #495: Use JSON RPC standard to return responses
The API connects clients with blockchain nodes using Pocket Network Protocol. 

Since blockchain nodes use JSON RPC, we should also use this standard to return responses that come from the API itself; such as the "Relays attempts exhasted" error.

Why?
Some of the clients (web3.js as example) expects a JSON RPC response for communication. If we are relaying the node's response its all good, however, if something unexpected happens and we return a message from the API as a plain JSON, web3.js will throw.

crisog:
Closes #495.

This PR adds a new dependency for JsonRpcErrors: https://github.com/teambition/jsonrpc-lite. It's mostly types, classes with static members; nothing too complex. The library itself has 0 dependencies so not a big deal for our dependency tree.


crisog:
I inverted the if logic in the `applicationRelay()` at `v1.controller.ts` — which is something I should have done in a separate PR. I can revert that change for a better review process.

src/controllers/v1.controller.ts
```diff
@@ -136,7 +137,7 @@ export class V1Controller {
         return this.loadBalancerRelay(redirect.loadBalancerID, rawData)
       }
     }
-    return new HttpErrors.InternalServerError('Invalid domain')
+    return jsonrpc.error(1, new jsonrpc.JsonRpcError('Invalid domain', -32000)) as ErrorObject
```
rem1niscence: why you use `jsonrpc.JsonRpcError` here and `ErrorObject` in other places, in what they differ?

crisog: jsonrpc.JsonRpcError 
```
{ 
message: string,
code: number
}
```

ErrorObject
```
jsonrpc: string,
id: number,
error: jsonrpc.JsonRpcError { 
message: string,
code: number
}
```

crisog: I think I got your confusion, using `jsonrpc.error` returns an ErrorObject — it's because in some places, I'm using ErrorObject constructor instead of `jsonrpc.error`.

rem1niscence: when one has to be used over the other?

src/controllers/v1.controller.ts
```diff
@@ -236,15 +237,17 @@ export class V1Controller {
 
       return await this.pocketRelayer.sendRelay(options)
     } catch (e) {
-      logger.log('error', e.message, {
+      const exception: ErrorObject = e
```
rem1niscence: I think more checks have to be done here, this is a general catch and I don't think all the errors that might come through will have the rpc format

src/controllers/v1.controller.ts
crisog: > 

No preference, both do the same — I'll just use one of them for consistency. The `jsonrpc.error()` is preferred.

rem1niscence: on that case can the other `ErrorObject` instances be replaced? for consistency

crisog: > 

Yes

src/controllers/v1.controller.ts
```diff
@@ -290,50 +293,54 @@ export class V1Controller {
...
+          keyPrefix: stickyKeyPrefix,
+          rpcID,
+          relaysLimit,
+          stickyOrigins,
+        },
       }
+
+      return await this.pocketRelayer.sendRelay(sendRelayOptions)
     } catch (e) {
+      const exception: ErrorObject = e
```
rem1niscence: same here

rem1niscence:
another point is that I saw you used the same -32000 for all the error responses, check the [JSON-RPC Specification](https://www.jsonrpc.org/specification#conventions) on section 5.1 for a more complete list of error objects to confirm is some of the ones set here can be more semantically correct

crisog:
> another point is that I saw you used the same -32000 for all the error responses, check the [JSON-RPC Specification](https://www.jsonrpc.org/specification#conventions) on section 5.1 for a more complete list of error objects to confirm is some of the ones set here can be more semantically correct

Sure, I took a look at that while I was writing the code but realized that none of those errors defined by the JSON-RPC standard define what we want, and using any of those would be misleading. Let's say you didn't find an entry in the DB, and you decide to return `-32601: Method not found` — the web3 client would think it sent the wrong method, which is clearly not the case... the error was on our side.

However, I agree that returning -32000 on everything would also be misleading. I suggest using the reserved range of server errors of the JSON-RPC specification (`-32000 to -32099`), and document the errors on our side.

rem1niscence:
all right, LGTM!

rem1niscence (approved):

crisog:
There's one test failing called `sends successful relay response as string` — I think this might as well be removed, or asserted on fail because we don't really want to return a string to the user at any time. The payload from calling a node should always be a JSON.

PR #537, opened 2021-12-17
Cherry picker median

nymd:


src/services/cherry-picker.ts
```diff
@@ -192,15 +180,11 @@ export class CherryPicker {
       await this.redis.lpush(blockchainID + '-' + id + '-relayTiming', elapsedTime.toFixed(3))
```
rem1niscence: can add a expire after this command so it doesn't stay dangling

rem1niscence (changes requested):

Issue #552, opened 2021-12-18
Cherry picker top bucket smearing


nymd:
The top bucket of the cherry picker should be more inclusive -- if you have an medianSuccessLatency of below a certain threshold, you should always be in the top bucket for random picking. That threshold should have a default of maybe 100ms but then be configurable in the database per chain.

nymd:
There are situations with the picker where the nodes' median success latency looks like this:

- 30ms
- 50ms
- 60ms
- 300ms
- 400ms

The top 3 nodes are obviously in the same city or very close to the API but the 3's response time is still double that of node 1. This is very significant in the picker's eyes and currently the 3 node will get much fewer relays.

However the difference in real world usage between 30ms and 60ms is minimal and nodes 1 through 3 should be equally rewarded for providing top tier service.

rem1niscence:
can this be closed?

PR #553, opened 2021-12-19
Cherry median refactor

nymd:
Changed from storing the relayTimingLog as a redis list to a JSON array.

Added a truncation feature that cuts the array size in half if it is > 500 entries.

Added a new constant `EXPECTED_SUCCESS_LATENCY` as the MVP for #552 
- If a node has a median success latency lower than this value, it is kept in the top bucket for random sorting

### DIFF
--- a/src/services/cherry-picker.ts
+++ b/src/services/cherry-picker.ts
@@ -278,7 +278,9 @@ export class CherryPicker {
   }
 
   reduceArray(raw: number[]): number[] {
-    // Remove every 2nd member of array
+    // Drop half of the array entries if it is over 500 elements.
+    // 500 is enough to get a good picture of the node's responses without making
+    // excessively long cache keys or arrays to sort.
     if (raw.length > 500) {
       const reducedArray = raw.filter(function (_, i) {
         return (i + 1) % 2
--- a/src/services/cherry-picker.ts
+++ b/src/services/cherry-picker.ts
@@ -373,20 +373,16 @@ export class CherryPicker {
     // be 10 times more likely to be selected than a node that has had failures.
     let weightFactor = 10
     let previousNodeLatency = 0
-    let latencyDifference = 0
 
     // This multiplier is tested to produce a curve that adequately punishes slow nodes
     const weightMultiplier = 15
 
     for (const sortedLog of sortedLogs) {
-      // Set the benchmark from the previous node and measure the delta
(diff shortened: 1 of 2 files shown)


src/services/cherry-picker.ts
```diff
@@ -253,23 +272,26 @@ export class CherryPicker {
...
-    // Convert redis items into floats
-    const convertedRaw = raw.map(function (x) {
-      return parseFloat(x)
-    })
-    const sorted = convertedRaw.sort((a, b) => a - b)
+  reduceArray(raw: number[]): number[] {
+    // Remove every 2nd member of array
+    if (raw.length > 500) {
+      const reducedArray = raw.filter(function (_, i) {
+        return (i + 1) % 2
```
Evalir: I think this is doing the opposite of what's intended? This is removing every _odd_ element of the array, instead of every even element.

Evalir: Also, what's the logic behind the 500? About the number, not why.

Evalir (approved):
Overall LGTM, just had a question:

src/services/cherry-picker.ts
nymd: Haha I don't really care if it's odd or even, I just want a reduction in array size while retaining the distribution.

500 seems like a large enough array of elapsed times to generate a good picture of a node's response while not making excessively long redis keys or arrays to sort.

Evalir: Ohh gotcha haha. Then yeah, this will work well, the comment just stood out to me. If we wanted to be "mathy/smart" about it we could also reduce the array by averaging adjacent values together, but I'm not necessarily sure this would perform better than just doing this :D.

Sounds good!

src/services/cherry-picker.ts
```diff
@@ -14,6 +14,11 @@ const TIMEOUT_LIMIT = 20
 // wont be exact
 const TIMEOUT_VARIANCE = 2
 
+// The maximum median latency for a node to be considered as providing optimal service.
+// This is temporarily a constant for MVP and will be moved to the database to be
+// variable per chain. Measured in seconds.
```
crisog, line 19: Good idea to have this per chain! 
:)

crisog (approved):
It looks good. I haven't touched cherry picker ever, so it was nice to take a look while reviewing!

PR #571, opened 2021-12-27
Return sync cache

rem1niscence:
adds boolean cache field to return on sync/chain checks


src/services/sync-checker.ts
```diff
@@ -48,23 +49,25 @@ export class SyncChecker {
     const syncedNodesKey = `sync-check-${sessionHash}`
     const syncedNodesCached = await this.redis.get(syncedNodesKey)
 
-    if (syncedNodesCached) {
+    const cached = !!syncedNodesCached
```
nymd: What is this nonsense double negative type conversion

nymd (approved):

src/services/sync-checker.ts
rem1niscence: welcome to js-land where nothing  works as expected, this double negative is no error, is just a way of trully converting a value that is "falsey" to a fully boolean, on this case, `syncedNodesCached` could come as `null` or string which are different falsey and not falsey datatypes respectively but can throw unexpected errors if treated as a fully boolean, with double exclamation mark we are forcing the null value to become a true boolean value. For more info check [this](https://stackoverflow.com/questions/29312123/how-does-the-double-exclamation-work-in-javascript)

Evalir: This is old stuff @rem1niscence . You can just do 
```suggestion
    const cached = Boolean(syncedNodesCached)
```

That answer is from 2015. Don't blame the language. :)

rem1niscence:
ez

PR #585, opened 2022-01-13
Gigastake v1

rem1niscence:
Adds basic gigastake functionality to load balancers (lb from now on) meaning that instead of lbs having their own apps, they'll share the same pool of apps from the biggest lb, this is mostly useful for lbs that have very to low usage that are keeping a high staked amount to themselves that is basically getting wasted.

### DIFF
--- a/src/controllers/v1.controller.ts
+++ b/src/controllers/v1.controller.ts
@@ -185,10 +185,26 @@ export class V1Controller {
     }
 
     try {
-      const loadBalancer = await this.fetchLoadBalancer(id, filter)
+      let loadBalancer = await this.fetchLoadBalancer(id, filter)
 
       if (!loadBalancer?.id) {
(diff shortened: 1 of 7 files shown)


src/controllers/v1.controller.ts
```diff
@@ -185,12 +185,47 @@ export class V1Controller {
...
+      } = {
+        gigastaked: loadBalancer.gigastakeRedirect || false,
+        lbApplication: undefined,
+      }
+
+      // Is this LB marked for gigastakeRedirect?
+      // Temporary: will be removed when live
+      if (gigastakeOptions.gigastaked) {
+        for (const redirect of JSON.parse(this.redirects)) {
+          if (this.host.toLowerCase().includes(redirect.blockchain)) {
```
Evalir: If I understand correctly, this is trying to find a `redirect` that has a certain `blockchain` property? Because that `break` on line 224 is a tad smelly haha. I think we can just use `find` on the redirects array.

nymd (approved):

PR #615, opened 2022-01-25
feat: refresh pocketjs instance at intervals

rem1niscence:
Adds a refresher at intervals of the pocketjs library


crisog:
Let those relays in! 🔥

nymd (approved):

PR #644, opened 2022-01-29
Stop pulling sessions with pocket-js

crisog:
Since we are now directly calling dispatchers, we no longer want to use pocket-js to pull nodes from session.

### DIFF
--- a/src/services/pocket-relayer.ts
+++ b/src/services/pocket-relayer.ts
@@ -2,16 +2,7 @@ import axios, { AxiosRequestConfig, Method } from 'axios'
 import { Redis } from 'ioredis'
 import jsonrpc, { ErrorObject, IParsedObject } from 'jsonrpc-lite'
 import { JSONObject } from '@loopback/context'
-import {
-  PocketAAT,
-  Session,
-  RelayResponse,
-  Pocket,
-  Configuration,
-  HTTPMethod,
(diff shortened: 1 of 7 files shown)


src/services/pocket-relayer.ts
```diff
@@ -599,46 +608,31 @@ export class PocketRelayer {
...
 
-    // Start the relay timer
-    const relayStart = process.hrtime()
+    const sessionHeader = new SessionHeader(appPublicKey, blockchainID, BigInt(0))
 
     // Converts the rpc response in a way that is compatible with pocketjs for
     // sending relays through
-    let nodes: Node[] = (sessionHeader.data.session.nodes as Node[]).map(
+    let nodes: Node[] = (dispatchResponse.data.session.nodes as Node[]).map(
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
rem1niscence: This mapping could be moved to a static function on the class or to `helpers.ts`

rem1niscence:
Also just so you not forget, the `pocketSession` variable after success is binded to a class variable and is used in several other places `sendRelay function, chain checker and sync checker` even though in those places is used for logging, that also needs to be removed to avoid `undefined` errors

rem1niscence:
LGTM!

PR #652, opened 2022-02-01
Pass dispatcher as string

rem1niscence:
Dispatchers injected through loopback as URL() list were having unexpected urls from node runners, this pass the urls as strings so they keep the intended value


crisog (approved):
stupid loopback, lgtm

nymd (approved):

PR #655, opened 2022-02-02
Revert to pocketjs

rem1niscence:
Revert the latests change made to use the pocket-rpc class for obtaining a session while keeping the latests features. This does not remove the pocket-rpc class at the moment just disconnects it from the normal flow.


src/config/pocket-config.ts
```diff
@@ -18,7 +18,7 @@ export type PocketConfiguration = {
 }
 
 export const DEFAULT_POCKET_CONFIG = {
-  maxDispatchers: 40,
+  maxDispatchers: 45,
```
nymd: 44

src/sequence.ts
```diff
@@ -92,4 +99,32 @@ export class GatewaySequence implements SequenceHandler {
...
+      ((await context.get('dispatchURL')) as string).split(',').map((dist) => new URL(dist))
+    )
+    const configuration: Configuration = await context.get('pocketConfiguration')
+    const clientPrivateKey: string = await context.get('clientPrivateKey')
+    const clientPassphrase: string = await context.get('clientPassphrase')
+
+    if (!(await redis.get(POCKET_JS_INSTANCE_TIMEOUT_KEY))) {
+      const pocket = await getPocketInstance(dispatchers, configuration, clientPrivateKey, clientPassphrase)
+
+      const nextInstanceRefresh = getRandomInt(POCKET_JS_TIMEOUT_MIN, POCKET_JS_TIMEOUT_MAX)
```
nymd, line 113: Wasn't this hitting dispatch too much?  At first we thought the refreshing helped. But then we found out loopback wasn't really refreshing pocket-js.

Once that was fixed, IIRC, it was getting way too many dispatches and causing load.

nymd (changes requested):

src/sequence.ts
rem1niscence: yeah on that last part we thought on increasing the time to refresh, won't call the function then but will leave it there for in the future wants to try a bigger threshold for instance reload

nymd: Yea I don't think it was a question of threshold; I think there was a logic issue. Let's go without it for now; just normal old pocket-js. If you want to try this on canary or a single region, we can do that.

rem1niscence: agree

PR #663, opened 2022-02-04
sync check logic fix

nymd:


src/services/sync-checker.ts
```diff
@@ -190,14 +190,26 @@ export class SyncChecker {
       })
     }
 
+    // Make sure nodes aren't running too far ahead of altruists
+    if (currentBlockHeight + syncCheckOptions.allowance > altruistBlockHeight) {
```
rem1niscence: this is a personal pet peeve but on comparisons like this I recommend to put parentheses on the operations that must be grouped, on this case the sum, as denotes better how should be performed regardless of the operations order of the language, even though is kinda obvious in this case

```suggestion
    if ((currentBlockHeight + syncCheckOptions.allowance) > altruistBlockHeight) {
```

src/services/sync-checker.ts
```diff
@@ -132,17 +132,17 @@ export class SyncChecker {
...
     }
 
-    // If there's at least 2 nodes, make sure at least two of them agree on current highest block to prevent one node
+    // If there's at least three nodes, make sure at least three of them agree on current highest block to prevent one node
     // from being wildly off
     if (
       !errorState &&
-      nodeSyncLogs.length >= 2 &&
+      nodeSyncLogs.length >= 3 &&
       nodeSyncLogs[0].blockHeight > nodeSyncLogs[1].blockHeight + syncCheckOptions.allowance
```
crisog: I think this needs to be updated to compare 3 nodeSyncLogs

rem1niscence (approved):

src/services/sync-checker.ts
crisog: Never mind, since we are sorting sync logs by height this is not required.

src/services/sync-checker.ts
```diff
@@ -628,9 +665,26 @@ export class SyncChecker {
   parseBlockFromPayload(payload: object, syncCheckResultKey: string): number {
     const rawHeight = payload[`${syncCheckResultKey}`] || '0'
 
-    const blockHeight = blockHexToDecimal(rawHeight)
+    return blockHexToDecimal(rawHeight)
+  }
+
+  // If all nodes reporting height above 0 are ahead altruist,
+  // we should trust the nodes.
+  isAltruistTrustworthy(altruistBlockHeight: number, nodeSyncLogs: NodeSyncLog[]): boolean {
```
rem1niscence: this function is redundant, you already have the nodeLogs sorted by blockheight, you could instead check if the last node's blockheight is greater than the altruist block height

crisog: Totally agree, before the last commit I didn't even think about this function as it was not needed but for some reason included it later. Thanks!

src/services/sync-checker.ts
```diff
@@ -159,6 +160,23 @@ export class SyncChecker {
     // Consult Altruist for sync source of truth
     const altruistBlockHeight = await this.getSyncFromAltruist(syncCheckOptions, blockchainSyncBackup)
 
+    // Determine whether altruist is trustworthy or not
+    const isAltruistTrustworthy = altruistBlockHeight >= highestNodeBlockHeight
```
rem1niscence: shouldn't be lowestNodeBlockHeight? on this case even if the other nodes before the highest one are lower than altruist still makes the altruist not trustworthy

PR #674, opened 2022-02-10
Remove exhausted sessions

rem1niscence:
specifically for load balancers, in case they have apps with exhausted sessions, will periodically remove them for a short period of time (10s) and try to attempt using another app from the limit if available, up to three attempts until giving up and simply picking a random app


tests/mocks/pocketjs.ts
```diff
@@ -134,6 +135,22 @@ export class PocketMock {
     this.relayCounter = {}
   }
 
+  /**
+   * Generates a node with random address/public key
+   * @returns node with random address/public key
+   */
+  static getRandomNote(): Node {
```
crisog, line 142: ~~Note~~ Node?

rem1niscence:
Closed ATM due to not being an immediate need

PR #676, opened 2022-02-11
Use pocketjs sessionkey

rem1niscence:
replaces the blockchainHash function for simply getting the pocket sessionKey and use it throughout the whole session


crisog (approved):
There are a good amount of tests around that too, looks good to me overall

PR #682, opened 2022-02-18
fix: unhandled fallback error responses

crisog:
We've got a handful of reports of users getting unformatted responses. The reason was that responses from altruists were not being JSON enforced as they should. 

The response from fallbacks (altruists) was being JSON-stringified, and that automatically makes it parseable to a JSON object despite it being a plain string (most of the cases HTML with bad responses from nodes).

This PR checks whether the given string passes an enforcement check without being stringified, to avoid any wrong responses reaching the clients. It also includes:

- Polishing of the error system, as well as updated tests for all of the changes made.
- Better naming for some errors in our system
- Clients can't send invalid JSON req bodies for load balancer relays.
- Error handling `rpcID` extraction from raw data.
- Service nodes responding with bad data returned a custom error response to the user saying that the response was invalid. No altruist attempts were being made; now fixed and it does.
- gatewayAAT fields don't throw anymore if some field is missing (mainly happened with 'version')
- nodejs assert for promises asserting (https://nodejs.org/api/assert.html)
- Supported node version to `16.14` as this is what we are using for Dockerfile `16-slim` which is a slim `16.14`.


crisog:
Link T-614

crisog:
Link T-576

nymd (approved):

PR #691, opened 2022-02-23
Testnet workflow

TakGN:


rem1niscence:
LGTM! Let's wait for the action to be deployed and do a manual check on it before merge

rem1niscence (approved):

PR #700, opened 2022-02-24
feat: change altruists timeout

rem1niscence:
Change altruists to use a default hardcoded value from the API


src/utils/constants.ts
```diff
@@ -14,3 +14,4 @@ export const WS_ONLY_METHODS = [
 export const POCKET_JS_INSTANCE_TIMEOUT_KEY = 'pocket-js-timeout'
 export const POCKET_JS_TIMEOUT_MIN = 60 // Seconds
 export const POCKET_JS_TIMEOUT_MAX = 120 // Seconds
+export const DEFAULT_ALTRUIST_TIMEOUT = 60000 // Millisseconds
```
crisog: sseconds 8)

rem1niscence: hands faster than brain 🚗

PR #712, opened 2022-03-02
feat: Add default blockchain path if required

rem1niscence:
Some blockchains require to add a path to the rpc url of the nodes all the time, people might not now this or think is not needed on a managed service like this one, for that a default path is added on such blockchains


crisog (approved):
Looks good to me! I'm pretty sure this could have been a big issue for regular users that copied our endpoint from the Portal and pasted it into Metamask; as you said, assuming that we are handling the path.

PR #713, opened 2022-03-08
fix: enable tls connection verification

crisog:
The TLS verification was disabled at Node level. This was causing the server to connect to make insecure connections to HTTP endpoints. Our service should enforce HTTPS.


crisog:
Already 3 days, seems like we can merge today

PR #720, opened 2022-03-11
[DO NOT MERGE] fix: exclusion of synced nodes by sync checker

crisog:
This PR introduces the following change:
- Sync check no longer favors altruists height as source of truth on a session with most of the nodes unsynced. This old behavior punished those nodes that were in sync, if altruist was also out of sync - as the majority of other nodes.


crisog:
I will be removing the first change on this PR, to have it on a separate one. I'll only leave this with the sync check modifications.

crisog:
We are going to use a different approach to solve this issue. Closing this PR.

PR #728, opened 2022-03-16
feat: altruist & redirect moved to DB

### LINKED ISSUE #721: Move Redirects and Altruists to db
Change hardcoded envs to db collections

crisog:
Closes #721.

The cache utility file will still require an ENV for its operation called `POKT_MAINNET_NODE` working as "altruist url".

The reason is that it's not worth adding all of the dependencies needed to fetch this from DB, even more if this is always a POKT node. It would make sense if this was a per-chain thing.


src/controllers/v1.controller.ts
```diff
@@ -130,16 +127,36 @@ export class V1Controller {
     })
     rawData: object
   ): Promise<string | ErrorObject> {
-    const parsedRawData = parseRawData(rawData)
```
crisog, line 133: This was not being handled and would throw if rawData couldn't be parsed

src/models/blockchains.model.ts
```diff
@@ -25,6 +25,27 @@ class SyncCheckOptions {
   allowance?: number
 }
 
+@model()
+class BlockchainRedirect {
+  @property({
+    type: 'string',
+    required: true,
+  })
+  alias: string
```
crisog, line 34: named this 'alias' instead of blockchain after I saw it being named differently on two redirects for the same chain

src/services/pocket-relayer.ts
```diff
@@ -200,7 +194,13 @@ export class PocketRelayer {
     }
 
     const data = JSON.stringify(parsedRawData)
-    const limitation = await this.enforceLimits(parsedRawData, blockchainID, requestID, logLimitBlocks)
+    const limitation = await this.enforceLimits(
+      parsedRawData,
+      blockchainID,
+      requestID,
+      logLimitBlocks,
+      blockchainAltruist
```
crisog, line 202: No longer obtainable app-wide, so must be passed now

src/utils/cache.ts
```diff
@@ -7,7 +7,7 @@ import { hashBlockchainNodes } from './helpers'
 
 const logger = require('../services/logger')
 
-const ALTRUIST_URL = JSON.parse(process.env.ALTRUISTS)?.['0001']
+const MAINNET_ALTRUIST_URL = String(process.env.POKT_MAINNET_NODE_URL)
```
crisog, line 10: The reason for still using an ENV here is explained in PR description

nymd (approved):

rem1niscence (approved):

PR #730, opened 2022-03-17
config: add testnet node env & run deployment artifacts

crisog:
There was only an ENV for a Pocket mainnet node, but we also have testnet. Added both to gh workflows


nymd (approved):

src/utils/cache.ts
```diff
@@ -7,7 +7,10 @@ import { hashBlockchainNodes } from './helpers'
 
 const logger = require('../services/logger')
 
-const MAINNET_ALTRUIST_URL = String(process.env.POKT_MAINNET_NODE_URL)
+const POCKET_NETWORK_NODE_URL =
+  String(process.env.REGION_NAME) === 'testnet'
```
rem1niscence: I mean why not a single env and change the value in both deployments?

crisog: > I mean why not a single env and change the value in both deployments?

True, just pocket network node.

crisog: > I mean why not a single env and change the value in both deployments?

Just realized this might be not be possible due to it being stored as a secret. It might be possible if I manually set the ENV for all deployments, but using shared env json file its not a thing.

PR #732, opened 2022-03-17
feat: get blockchain by domain

crisog:
The problem PR #728 was that it failed to fetch the blockchain when a simple URL hit. There wasn't a way to find a blockchain just by the domain.

This PR introduces a way to do this, tested locally and also in canary. It works like a charm. All of the errors for invalid domain disappeared.


crisog:
Fixing the test asap

nymd (approved):

PR #735, opened 2022-03-20
Redirect apps to gigastake

rem1niscence:
Redirects all the apps to a load balancer gigastake if the chain has one, doesn't matter whether the app is staked or not


nymd (approved):

rem1niscence:
Bad branch origin

PR #741, opened 2022-03-21
Add evm error

rem1niscence:
Adds evm error and refactors a bit for when the need to add non evm error arise

### DIFF
--- a/src/utils/errors.ts
+++ b/src/utils/errors.ts
@@ -4,7 +4,7 @@ export const EVM_ERROR_CODES = [
   // JSON RPC Standard errors
   '-32',
   // Execution reverted error
-  '-3',
+  '3',
 ]
 
 // EVM clients rely on JsonRpc standard, so we check for those errors.
--- a/src/utils/errors.ts
+++ b/src/utils/errors.ts
@@ -3,14 +3,20 @@ import { parseJSONRPCError } from './parsing'
 export const EVM_ERROR_CODES = [
   // JSON RPC Standard errors
   '-32',
+  // Execution reverted error
+  '-3',
 ]
 
 // EVM clients rely on JsonRpc standard, so we check for those errors.
 export function isEVMError(response: string): boolean {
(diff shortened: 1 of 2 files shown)


src/utils/errors.ts
```diff
@@ -3,14 +3,20 @@ import { parseJSONRPCError } from './parsing'
 export const EVM_ERROR_CODES = [
   // JSON RPC Standard errors
   '-32',
+  // Execution reverted error
+  '-3',
```
nymd: doesn't this automatically include the line above?

crisog (approved):
LGTM

src/utils/errors.ts
rem1niscence: no as it checks for exact matches

crisog: Yes, kind of a coincidence though. There are also custom codes 1 and wouldn't match -32.

crisog: > no as it checks for exact matches

Why not? It's `.includes()`

crisog: The error should be 3 and not -3 btw

rem1niscence: yep, my bad I got confused, on that case is not working as expected, back to draft

rem1niscence: > The error should be 3 and not -3 btw

oh that's why yes

src/utils/errors.ts
```diff
@@ -3,14 +3,20 @@ import { parseJSONRPCError } from './parsing'
...
 export function isEVMError(response: string): boolean {
+  return isError(response, EVM_ERROR_CODES)
+}
+
+function isError(response: string, errorsToCheck: string[]): boolean {
   try {
     const { code } = parseJSONRPCError(response)
 
-    return EVM_ERROR_CODES.some((error) => String(code).includes(error))
+    return errorsToCheck.some((error) => String(code).includes(error))
```
crisog, line 19: I think we should remove the string cast here, so we can have exact matches

PR #756, opened 2022-03-25
feat: Contract address/method allowlist for endpoint security

### LINKED ISSUE #695: [FEATURE REQUEST] Contract/RPC Method Whitelisting for Endpoint Security.
**Feature Overview**

The current security measures of the Portal are really effective at giving developers building within traditional web2 systems the power they need to protect their endpoint from free-riding. However, these current restrictions don't offer any protection to developers hoping to publish their application interfaces to decentralized immutable/archival storage platforms like IPFS or arweave. 

In the case of these developers, if they wish to use an RPC endpoint directly in their application interface, they need to be open to the fact that anyone in theory would be able to use that endpoint. If the Portal would allow developers to whitelist specific RPC methods, possibly in aggregate by contract address, then endpoint free-riding is no longer a problem. As long as application developers restrict access to only the methods called in their interface, the only thing someone could do if they decided to free-ride the endpoint is interact the same way the interface was designed to.

**Solution Suggestion**

An ability to restrict RPC interactions to a whitelist of:

1. Contract addresses
2. Specific RPC methods

**Additional Context**

Using an EVM compatible chain for reference, these are the `eth_` methods where a contract address would be coming through as a parameter (or encoded in data). It's very likely that not all of these methods would need to be considered when whitelisting by contract, this just captures the context of what exactly might need to be considered to filter by contract address.

Contract method specific read and writes are primarily going to come through the two following methods, and would need to be parsed out to then be filtered if whitelisting by contract address: 

1. `eth_sendRawTransaction`: The transaction would need to be decoded, and the `to` field is where the address could be checked against a contract whitelist. 
2. `eth_call`: The `to` field of the request could be checked against a contract whitelist.

Additionally, the following methods provide information about the contract that has the potential to be useful when filtered to a contract level:

1. `eth_getBalance`: If the contract itself holds a native token balance, the first parameter of this call would be the address to check against the whitelist.
2. `eth_getTransactionCount`: Typically used for automatically getting the account nonce when constructing transactions. Not sure of how often it is used in a contract specific manner client side. Like above, the first parameter of this call would be the address to check against the whitelist.

Finally the following methods also interact on a contract specific basis, but aren't used in the typical read/write interactions:

1. `eth_getStorageAt`: Typically storage values are accessed through getters via `eth_call`, so likely not used. First parameter is the address.
2. `eth_getLogs`: Address would be under the `address` field of object used for the parameters.
3. `eth_getCode`: First parameter would be the address.

crisog:
This PR adds the contract address and method whitelisting featured requested on #695.

Two fields have been incorporated in the database, as optional fields.
- `whitelistMethods` which is an array of objects that contain both blockchainID and the methods.
- `whitelistContracts` which is an array of objects that contain both blockchainID and the contract addresses.

The feature supports both individual and bundled RPC calls. A good amount of tests have been added covering most of the cases around it.

The term "Limitation" has been replaced by "Restriction".

Closes #695.


crisog:
Close T-2556

src/utils/constants.ts
```diff
@@ -11,6 +11,26 @@ export const WS_ONLY_METHODS = [
   'newBlockFilter',
 ]
 
+export const EVM_CHAINS = [
```
nymd: Maybe AVAX should be in there as it has C-chain?

But more importantly, I think this list will need to be shared with the UI so that it can expose this as a feature. So, DB?

src/utils/evm/restrictions.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export async function enforceEVMRestrictions(
+  application: Applications,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  parsedRawData: Record<string, any>,
+  blockchainID: string,
+  requestID: string,
+  rpcID: number,
+  logLimitBlocks: number,
+  altruistUrl: string
```
nymd: Nitpicky: altruistURL

nymd (changes requested):
Does this unwrap bundled relays that have more than 1 evm call inside []? I couldn't tell.

crisog:
> Does this unwrap bundled relays that have more than 1 evm call inside []? I couldn't tell.

Yes it does

src/utils/constants.ts
crisog: > Maybe AVAX should be in there as it has C-chain?
> 
> But more importantly, I think this list will need to be shared with the UI so that it can expose this as a feature. So, DB?

Definitely. I think we could add a boolean called `evm` to the blockchain collection.

src/utils/evm/restrictions.ts
crisog: > Nitpicky: altruistURL

sure, ser

src/models/applications.model.ts
```diff
@@ -111,6 +111,18 @@ type GatewayAAT = {
 export type GatewaySettings = {
   whitelistOrigins: string[]
   whitelistUserAgents: string[]
+  whitelistContracts?: WhitelistContract[]
+  whitelistMethods?: WhitelistMethod[]
```
Evalir, line 115: This is an overall typo we have on these options, they should be `whitelistedContracts/Methods`, not `whitelist` haha

src/utils/evm/parsing.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  if (method === 'eth_sendRawTransaction') {
+    // firstParameter is the raw tx hex
+    return decodeEthRawTxAddress(firstParameter)
+  } else if (method === 'eth_call') {
+    return firstParameter?.to
+  } else if (method === 'eth_getLogs') {
+    return firstParameter?.address
+  } else if (
+    method === 'eth_getCode' ||
```
rem1niscence: add this to an array and check with `includes`

src/utils/evm/parsing.ts
```diff
@@ -0,0 +1,44 @@
+import { ethers } from 'ethers'
+import { parseMethod } from '../parsing'
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function extractContractAddress(rawData: Record<string, any>) {
+  const [firstParameter] = rawData.params
+  const method = parseMethod(rawData)
+
+  if (method === 'eth_sendRawTransaction') {
```
rem1niscence: you could change this to a switch

src/utils/evm/parsing.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  // If the method is not supported, return undefined
+  return undefined
+}
+
+export function decodeEthRawTxAddress(rawTxBytes: string) {
+  const decodedTx = ethers.utils.RLP.decode(rawTxBytes)
+
+  const [
+    ,
```
rem1niscence, line 33: I supposed those comments are per field, why not between the commas? syntax error?

src/utils/evm/restrictions.ts
```diff
@@ -0,0 +1,47 @@
...
+  parsedRawData: Record<string, any>,
+  blockchainID: string,
+  requestID: string,
+  rpcID: number,
+  logLimitBlocks: number,
+  altruistURL: string
+): Promise<ErrorObject | void> {
+  const method = parseMethod(parsedRawData)
+
+  if (WS_ONLY_METHODS.includes(method)) {
```
rem1niscence, line 20: these ifs end with return so no `else if` is necessary

src/utils/evm/whitelist.ts
```diff
@@ -0,0 +1,25 @@
...
+    return jsonrpc.error(rpcID, new jsonrpc.JsonRpcError('Restricted endpoint: method not allowed.', 0)) as ErrorObject
+  }
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function enforceContractWhitelist(rpcID: number, rawData: Record<string, any>, whitelistedContracts: string[]) {
+  const contractAddress = extractContractAddress(rawData)
+
+  if (!contractAddress) {
+    return
```
rem1niscence: inconsistent return, plz add the return type to the function signature and enforce the returns to follow that

rem1niscence (changes requested):

src/utils/evm/parsing.ts
```diff
@@ -0,0 +1,44 @@
+import { ethers } from 'ethers'
+import { parseMethod } from '../parsing'
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function extractContractAddress(rawData: Record<string, any>) {
```
Evalir: Just a note on consistency: the functions from here on don't have return types.

src/utils/evm/parsing.ts
crisog: > I supposed those comments are per field, why not between the comas? syntax error?

Yea I didn't notice this one, it should've been the linter fixing it.

src/utils/evm/restrictions.ts
```diff
@@ -0,0 +1,47 @@
...
+      rpcID,
+      new jsonrpc.JsonRpcError(
+        `${method} method cannot be served over HTTPS. WebSockets are not supported at the moment.`,
+        -32053
+      )
+    ) as ErrorObject
+  } else if (application?.gatewaySettings?.whitelistMethods?.length > 0) {
+    const restriction = application.gatewaySettings.whitelistMethods.find((x) => x.blockchainID === blockchainID)
+
+    if (!restriction) {
```
Evalir: I think the logic here might be wrong... These ifs are not falling through. If I have, for example, whitelisted methods, and this check passes, it will return, even though I could've had whitelisted contracts that would've failed the check. Either we have that, or I'm not following correctly.

I'd code this as separate `if`s, and just return the error object, if there's any.

src/utils/evm/restrictions.ts
```diff
@@ -0,0 +1,47 @@
...
+export async function enforceEVMRestrictions(
+  application: Applications,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  parsedRawData: Record<string, any>,
+  blockchainID: string,
+  requestID: string,
+  rpcID: number,
+  logLimitBlocks: number,
+  altruistURL: string
+): Promise<ErrorObject | void> {
```
Evalir: In general returning `undefined` if nothing's wrong feels a tad wrong haha... I'm not gonna bikeshed or nit too much because we're moving to go very very soon, but this type of function can definitely let things fall through the cracks unnoticed. It's better to return some kind of Symbol, string, or a boolean.

package.json
```diff
@@ -84,6 +84,7 @@
     "dd-trace": "^2.0.0",
     "dotenv": "^15.0.0",
     "eslint-plugin-import": "^2.23.4",
+    "ethers": "^5.6.1",
```
Evalir, line 87: Note that if we only need the `utils` package we can install that directly instead of using the whole ethers package. It's still unimportant though as it's not being sent to users.

Evalir (commented):
Overall my main concern is with the `enforceEVMRestrictions` function—I'm not sure it's coded properly.

src/utils/evm/parsing.ts
Evalir: Why add it to an array? Feels like an extra step that wastes CPU cycles.

src/utils/evm/whitelist.ts
Evalir: What do you mean with "inconsistent return"

src/models/applications.model.ts
crisog: > This is an overall typo we have on these options, they should be `whitelistedContracts/Methods`, not `whitelist` haha

Just followed naming of other fields, such as `whitelistOrigins`. I was aware that whitelisted was the appropriate naming but would require others to be changed too.

Evalir: Yeah for sure, no need to take action on it tbh—we'll fix all of these once we move over to Golang.

crisog: Not require, just felt that it would need to be a general change

src/utils/evm/restrictions.ts
crisog: You are right, made it so it only returns if there's an error.

src/utils/evm/parsing.ts
rem1niscence: code readability

src/utils/evm/whitelist.ts
rem1niscence: one empty return and the other return and error

src/utils/evm/parsing.ts
Evalir: Chained `or`s are pretty readable? I really don't think it's worth sacrificing CPU cycles for this.

rem1niscence: I do, not my PR so the owner can do it if they want, just stated my opinion

src/utils/evm/parsing.ts
crisog: > you could change this to a switch

Valid! I did first do it with a switch, for some reason I did not use it because the way I was parsing the methods. That got replaced shortly after and I didn't turn this back to a switch.

Replaced now!

src/utils/evm/parsing.ts
crisog: > I do, not my PR so the owner can do it if they want, just stated my opinion

I moved this to a switch, more readable there :)

src/utils/evm/whitelist.ts
crisog: These functions have been re-done, resolving this.

rem1niscence (approved):

PR #758, opened 2022-03-25
[WIP] feat: implement cache manager

crisog:
We are relying too much on remote cache. This adds a layer of local caching to avoid doing excessive round trips to external redis instances.

Two packages are added:
2. https://www.npmjs.com/package/cache-manager
2. https://www.npmjs.com/package/cache-manager-ioredis


crisog:
This has been implemented at #770.

PR #759, opened 2022-03-26
72 redis calls reduced to 3

nymd:


src/services/cherry-picker.ts
```diff
@@ -156,6 +161,26 @@ export class CherryPicker {
     return node
   }
 
+  // Fetch app/node's service log from redis
+  async fetchRawServiceLogs(blockchain: string, rawNodeIDs: string[]): Promise<{ [id: string]: string }> {
+    const redisKeys = [] as string[]
```
rem1niscence:
```suggestion
    const redisKeys : string[] =  []
```

src/services/cherry-picker.ts
```diff
@@ -156,6 +161,26 @@ export class CherryPicker {
...
+    const rawServiceLogs = {} as { [id: string]: string }
```
rem1niscence:
```suggestion
    const rawServiceLogs: { [id: string]: string } = {} 
```

src/services/cherry-picker.ts
```diff
@@ -156,6 +161,26 @@ export class CherryPicker {
...
+
+    rawNodeIDs.forEach(async (rawNodeID) => {
+      redisKeys.push(blockchain + '-' + rawNodeID + '-service')
+    })
+
+    const rawRedisLogs = await this.redis.mget(redisKeys)
```
rem1niscence, line 125: `mget` in the end is still O(N) so I wouldn't say they're completely reduced but, definitely doing it in one single call will be faster due to doing a single trip.

src/services/cherry-picker.ts
rem1niscence: `as` is better when you're casting from one type to another, on this case as you're defining the type is better to simply put it at the definition

src/services/cherry-picker.ts
rem1niscence: same here

src/services/cherry-picker.ts
```diff
@@ -113,11 +71,11 @@ export class CherryPicker {
     /*
     RE-ENABLE LOGS to examine cherry picker behaviour
     */
-    // logger.log('info', 'CHERRY PICKER STATS Sorted logs: ' + JSON.stringify(sortedLogs), {
-    //   requestID: requestID,
-    //   blockchainID: blockchain,
-    //   sessionKey: sessionKey,
-    // })
+    logger.log('info', 'CHERRY PICKER STATS Sorted logs: ' + JSON.stringify(sortedLogs), {
```
rem1niscence, line 74: later let's just add a debug flag to this to avoid changing so much, that same flag could even be used in other places too

rem1niscence (changes requested):

src/services/cherry-picker.ts
nymd: line was removed

src/services/cherry-picker.ts
nymd: Yea true I'm assuming the problem is the network traffic and not the actual calls on redis

src/services/cherry-picker.ts
nymd: agreed

src/services/cherry-picker.ts
```diff
@@ -102,9 +46,23 @@ export class CherryPicker {
     for (const node of nodes) {
       rawNodes[node.publicKey] = node
       rawNodeIDs.push(node.publicKey)
-      const rawServiceLog = await this.fetchRawServiceLog(blockchain, node.publicKey)
+    }
+
+    // Pull all service & failure & error logs
+    const rawServiceLogs = await this.fetchRawLogs(blockchain, rawNodeIDs, 'service')
+    const rawFailureLogs = await this.fetchRawLogs(blockchain, rawNodeIDs, 'failure')
+    const rawErrorLogs = await this.fetchRawLogs(blockchain, rawNodeIDs, 'errors')
```
Evalir, line 54: We can probably optimize this by fetching them all at the same time rather than sequentially, with `Promise.all`

src/services/cherry-picker.ts
```diff
@@ -102,9 +46,23 @@ export class CherryPicker {
...
 
-      sortedLogs.push(await this.createUnsortedLog(node.publicKey, blockchain, rawServiceLog!))
+    for (const node of nodes) {
```
Evalir, line 56: Does order/sequential processing matter here? Wonder if we can also optimize this by also mapping over these with a `Promise.all`

rem1niscence (approved):

src/services/cherry-picker.ts
```diff
@@ -156,6 +114,24 @@ export class CherryPicker {
...
+    const rawServiceLogs = {} as { [id: string]: string }
+
+    const redisKeys = rawNodeIDs.map(function (rawNodeID) {
+      return `${blockchain}-${rawNodeID}-${logType}`
+    })
+
+    const rawRedisLogs = await this.redis.mget(redisKeys)
+
+    let logCount = 0
+    rawNodeIDs.forEach(async (rawNodeID) => {
```
Evalir: I think we don't need the `async` here
```suggestion
    rawNodeIDs.forEach(rawNodeID => {
```

Evalir (commented):
A few low hanging fruits in terms of optimization, but this is looking good :)

src/services/cherry-picker.ts
nymd: order does not matter at that point. would be a good code optimization after, but not messing w/ it now.

src/services/cherry-picker.ts
nymd: Could even be combined into just one trip to redis but since it's Saturday, this is a fine way to test if this is the actual issue

PR #771, opened 2022-03-28
cherry algo improvements for 24 nodes per session

nymd:
- fixed bug where if there weren't any nodes below the `EXPECTED_SUCCESS_LATENCY` the picker was using an old value as the first node's benchmark
- changed `EXPECTED_SUCCESS_LATENCY` to 150ms to allow more nodes in the top bucket and this value will keep us competitive with centralized services without much variance if all nodes can meet this expectation
- changed algo's `WEIGHT_MULTIPLIER` to avoid nodes that are far out from the expected latency
- changed benchmark so that if the previous node's latency is below expected_success then this node is measured against expected_success and not the previous node

### DIFF
--- a/src/services/cherry-picker.ts
+++ b/src/services/cherry-picker.ts
@@ -400,7 +400,7 @@ export class CherryPicker {
           }
         }
       }
-      // console.log(`${weightFactor}: attempts: ${sortedLog.attempts} success: ${Math.round(sortedLog.successRate * 100)}% median: ${sortedLog.medianSuccessLatency}s weighted: ${sortedLog.weightedSuccessLatency} ${sortedLog.id}`)
+
       // Set the benchmark for the next node
       previousNodeLatency = sortedLog.weightedSuccessLatency
     }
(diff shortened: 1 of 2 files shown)


src/services/cherry-picker.ts
```diff
@@ -392,10 +400,11 @@ export class CherryPicker {
           }
         }
       }
-
+      // console.log(`${weightFactor}: attempts: ${sortedLog.attempts} success: ${Math.round(sortedLog.successRate * 100)}% median: ${sortedLog.medianSuccessLatency}s weighted: ${sortedLog.weightedSuccessLatency} ${sortedLog.id}`)
```
rem1niscence: is this meant to stay?

nymd: Yea it made checking the results easy. it's commented out.

rem1niscence: if you also want it to be shown on prod, will need to change it to `logger.log` unless you only want it for local testing

nymd: it's commented out, it is just for me to use locally later. i'll delete it in the code.

rem1niscence (approved):

PR #772, opened 2022-03-28
batching redis mgets

nymd:


src/services/cherry-picker.ts
```diff
@@ -119,21 +117,38 @@ export class CherryPicker {
...
+      return `{${blockchain}}-${rawNodeID}-service`
+    })
+    const redisFailureKeys = rawNodeIDs.map(function (rawNodeID) {
+      return `{${blockchain}}-${rawNodeID}-failure`
+    })
+    const redisErrorKeys = rawNodeIDs.map(function (rawNodeID) {
+      return `{${blockchain}}-${rawNodeID}-errors`
     })
 
+    const redisKeys = redisServiceKeys.concat(redisFailureKeys).concat(redisErrorKeys)
```
crisog, line 135: could've used spread operator but works the same

crisog (approved):
Looks good!

src/services/cherry-picker.ts
rem1niscence, line 135:
```suggestion
    const redisKeys = [...redisFailureKeys, ...redisErrorKeys]
```

PR #777, opened 2022-03-29
Local cache improvements

rem1niscence:
local-cache has a `checkPeriod` variable which is approximately how many seconds it wait before checking and deleting all expired keys, lower the amount of that and enable `useClones` which is to store/retrieve cache variable references instead of copies for perfomance


rem1niscence:
Bad base branch

PR #778, opened 2022-03-29
feat: add clones and reduce check period for local cache

rem1niscence:
local-cache has a checkPeriod variable which is approximately how many seconds it wait before checking and deleting all expired keys, lower the amount of that and enable useClones which is to store/retrieve cache variable references instead of copies for perfomance

### DIFF
--- a/src/services/cache.ts
+++ b/src/services/cache.ts
@@ -9,19 +9,19 @@ export class Cache {
   constructor(redis: Redis) {
     this.redis = redis
     this.local = new NodeCache({
-      checkperiod: 60,
-      useClones: true,
+      checkperiod: 10,
+      useClones: false,
       deleteOnExpire: true,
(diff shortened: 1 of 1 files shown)


crisog:
> local-cache has a checkPeriod variable which is approximately how many seconds it wait before checking and deleting all expired keys, lower the amount of that and enable useClones which is to store/retrieve cache variable references instead of copies for perfomance

I think you meant disable useClones?

crisog (approved):

rem1niscence:
> > local-cache has a checkPeriod variable which is approximately how many seconds it wait before checking and deleting all expired keys, lower the amount of that and enable useClones which is to store/retrieve cache variable references instead of copies for perfomance
> 
> I think you meant disable useClones?

yeah that

PR #781, opened 2022-03-29
Local-cache-ttl-fix

rem1niscence:
Fixes error where local-cache was saving wrong tttl values

### DIFF
--- a/src/services/cache.ts
+++ b/src/services/cache.ts
@@ -136,6 +136,6 @@ export class Cache {
   private getLocalTTL(key: string) {
     const localTTL = this.local.getTtl(key) || 0
     // Gets time difference in seconds
-    return localTTL <= 0 ? localTTL : (localTTL - new Date().getTime()) / 1000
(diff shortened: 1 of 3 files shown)


crisog (approved):

src/services/cache.ts
```diff
@@ -130,9 +130,12 @@ export class Cache {
...
   }
 
+  // node-cache returns a timestamp of when the time is going to expire, while
+  // redis returns an integer meaning remaining seconds, this converts the
+  // node-cache ttl result to remaining seconds
   private getLocalTTL(key: string) {
     const localTTL = this.local.getTtl(key) || 0
     // Gets time difference in seconds
-    return localTTL > 0 ? localTTL : (localTTL - new Date().getTime()) / 1000
+    return localTTL <= 0 ? localTTL : (localTTL - new Date().getTime()) / 1000
```
crisog: even tho it looks better like this
```suggestion
    return localTTL > 0 ? (localTTL - new Date().getTime()) / 1000 : localTTL
```

PR #793, opened 2022-03-30
Local cache TTL factor 

rem1niscence:
Enhance the local ttl by adding a custom factor based in the remote ttl given

### DIFF
--- a/src/services/cache.ts
+++ b/src/services/cache.ts
@@ -13,12 +13,12 @@ export class Cache {
     this.ttlFactor = ttlFactor
   }
 
-  getLocalTTlL(ttl: number) {
+  getLocalTTL(ttl: number) {
     return Math.ceil(this.ttlFactor * ttl)
   }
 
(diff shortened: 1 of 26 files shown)


src/services/cache.ts
```diff
@@ -55,7 +62,7 @@ export class Cache {
   }
 
   async expire(key: string, ttlSeconds: number) {
-    await this.local.expire(key, ttlSeconds)
+    await this.local.expire(key, this.getLocalTTlL(ttlSeconds))
```
nymd: there's an extra 1 in this function name

nymd (changes requested):

src/services/cache.ts
rem1niscence: l you mean haha but you're right, on it

nymd (approved):
did you set the prod secret

rem1niscence:
> did you set the prod secret

yes

PR #806, opened 2022-03-31
fix: forward blockchain path for simple redirect

crisog:
This allows blockchains with a path, to be used with DNS redirects.

### DIFF
--- a/src/controllers/v1.controller.ts
+++ b/src/controllers/v1.controller.ts
@@ -144,7 +144,7 @@ export class V1Controller {
       // TODO: Move URL to ENV
       this.host = `${blockchainAliases[0]}.gateway.pokt.network`
 
-      const { blockchainRedirects } = await loadBlockchain(
+      const { blockchainRedirects, blockchainPath } = await loadBlockchain(
         this.host,
         this.cache,
         this.blockchainsRepository,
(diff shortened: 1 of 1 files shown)


src/controllers/v1.controller.ts
```diff
@@ -144,7 +144,7 @@ export class V1Controller {
       // TODO: Move URL to ENV
       this.host = `${blockchainAliases[0]}.gateway.pokt.network`
 
-      const { blockchainRedirects } = await loadBlockchain(
+      const { blockchainRedirects, blockchainPath } = await loadBlockchain(
```
nymd, line 147: This should get a change in the blockchains.model too yea?

crisog:
Close T-2716

src/controllers/v1.controller.ts
crisog: We moved the path to sync check, but it's no longer useful ONLY for the sync checks. We should go back to using the field outside of `syncCheckOptions`

crisog: I added `path` field to all of the blockchains that had a path in the DB already.

crisog: > This should get a change in the blockchains.model too yea?

Also, the `path` field was never removed from the model so no need to change.

nymd (approved):

PR #808, opened 2022-04-01
[develop] fix: eth_getLogs restriction failures & invalid node response

crisog:
This PR fixes two things:

1. As of now, if our altruist were not reachable during the eth_getLogs restrictions, we would propagate an error to the user and have them not use 'latest' as an option. 

This behavior is quite overkill for the purpose of this feature, and if our altruists are having a bad time then all of the systems relaying on us, and using 'latest' would take a hit until they are back up.

Solution: if we altruist is unreachable, ignore this restriction and let the relay pass with 'latest' flag. The worse that can happen is having a request from 0 to latest, but that's restricted if our altruist is up. 

2. Invalid node responses were being propagated to the user instead of being exclusively logged and fallback to an altruist. There's no point on propagating this error to the user, it's better to only log this for our use and let an altruist serve it.

### DIFF
--- a/src/services/pocket-relayer.ts
+++ b/src/services/pocket-relayer.ts
@@ -271,10 +271,7 @@ export class PocketRelayer {
             // If the parsing goes wrong, we get a response with 'invalid' type and the following message.
             // We could get 'invalid' and not a parse error, hence we check both.
             if (parsedRelayResponse.type === 'invalid' && parsedRelayResponse.payload.message === 'Parse error') {
-              throw new ErrorObject(
-                rpcID,
-                new jsonrpc.JsonRpcError('Service Node returned an invalid response', -32065)
-              )
+              throw new Error('Service Node returned an invalid response')
             }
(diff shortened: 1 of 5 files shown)


crisog:
Close T-2805

crisog:
Close T-2801

src/services/pocket-relayer.ts
```diff
@@ -271,10 +271,7 @@ export class PocketRelayer {
             // If the parsing goes wrong, we get a response with 'invalid' type and the following message.
             // We could get 'invalid' and not a parse error, hence we check both.
             if (parsedRelayResponse.type === 'invalid' && parsedRelayResponse.payload.message === 'Parse error') {
-              throw new ErrorObject(
-                rpcID,
-                new jsonrpc.JsonRpcError('Service Node returned an invalid response', -32065)
```
crisog, line 276: This was preventing these reqs from getting a fallback response.

nymd (approved):

PR #818, opened 2022-04-07
feat: blockchain allowlist

crisog:
Using a DB field to determine whether the user has restricted access to a specific blockchain or not

### DIFF
--- a/src/models/applications.model.ts
+++ b/src/models/applications.model.ts
@@ -111,7 +111,7 @@ type GatewayAAT = {
 export type GatewaySettings = {
   whitelistOrigins: string[]
   whitelistUserAgents: string[]
-  whitelistBlockchains: string[]
+  whitelistBlockchains?: string[]
   whitelistContracts?: WhitelistContract[]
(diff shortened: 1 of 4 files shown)


src/models/applications.model.ts
```diff
@@ -111,6 +111,7 @@ type GatewayAAT = {
 export type GatewaySettings = {
   whitelistOrigins: string[]
   whitelistUserAgents: string[]
+  whitelistBlockchains: string[]
```
nymd: Why are the two below optional but not this one? shouldn't this one be as well?

crisog: > Why are the two below optional but not this one? shouldn't this one be as well?

Yes

nymd:
Nice simple execution of the feature; I'm sure UI's part will be a bit harder haha

crisog:
> Nice simple execution of the feature; I'm sure UI's part will be a bit harder haha

A bit harder, but Albin also has a good setup to add it in a simple way

rem1niscence (approved):

PR #820, opened 2022-04-08
feat: add type id facet to log

crisog:


rem1niscence (approved):
the mind behind the idea of this change must be sooooooooo cooooooolllll

Evalir:
cringe brah

PR #825, opened 2022-04-14
Log first 80 characters of successful relay responses in datadog

blockjoe:
I've been unable to establish any patterns or ideas of where the occasional 'leaky' error that ends up being relayed to users has been coming from no matter how I slice the data. I also have struggled to be able to trace back any user report of these errors to the relay which it happened.

Being able to see the first few characters of successful responses would allow for being able to quickly identify the log in which it happened to then trace back the conditions in which it happened, at least in the case of the occasional default nginx bad gateway response that occasionally is surfaced.


blockjoe:
It would actually be now the first 200 characters of altruist fallbacks that gets logged.

crisog:
> It would actually be now the first 200 characters of altruist fallbacks that gets logged.

The reasoning behind this is to determine whether the connection errors reported by users are coming from our altruists or not. It can be turned off later

crisog (approved):

PR #829, opened 2022-05-06
Cherry-picker-changes

rem1niscence:
Several changes on the cherry picker service, nothing affects they way the results are calculated. Some of them include

* Fix typos and old comments pointing out to info not accurate anymore
* Remove unused fields on methods and places were used to send them

Fixed the error where the checks errors were taking into account for the cherry picker results.

And at last added in metrics recorder a ephemeral store of node relay successes to be consumed externally

### DIFF
--- a/src/services/cherry-picker.ts
+++ b/src/services/cherry-picker.ts
@@ -6,6 +6,8 @@ import { Cache } from './cache'
 
 const logger = require('../services/logger')
 
+const logStats = (process.env['LOG_CHERRY_PICKER_STATS'] || '').toLowerCase()
+
 // Amount of times a node is allowed to fail due to misconfigured timeout before
 // being removed from the session
 const TIMEOUT_LIMIT = 20
@@ -36,7 +38,13 @@ export class CherryPicker {
   // Record the latency and success rate of each node, 1 hor TTL
   // When selecting a node, pull the stats for each node in the session
   // Rank and weight them for node choice.
-  async cherryPickNode(application: Applications, nodes: Node[], blockchain: string, requestID: string): Promise<Node> {
+  async cherryPickNode(
+    application: Applications,
+    nodes: Node[],
+    blockchain: string,
+    requestID: string,
+    sessionKey: string
+  ): Promise<Node> {
     const rawNodes = {} as { [nodePublicKey: string]: Node }
     const rawNodeIDs = [] as string[]
     let sortedLogs = [] as ServiceLog[]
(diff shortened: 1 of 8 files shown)


src/services/cherry-picker.ts
```diff
@@ -6,6 +6,8 @@ import { Cache } from './cache'
 
 const logger = require('../services/logger')
 
+const logStats = (process.env['LOG_CHERRY_PICKER_STATS'] || '').toLowerCase()
```
Evalir, line 9: Q: Do we maybe wanna log something if this env var is not present?

src/services/cherry-picker.ts
```diff
@@ -68,22 +70,20 @@ export class CherryPicker {
...
+        blockchainID: blockchain,
+        sessionKey: sessionKey,
+      })
+    }
 
     // Iterate through sorted logs and form in to a weighted list
     let rankedItems = await this.rankItems(blockchain, sortedLogs, 50)
 
-    // If we have no nodes left because all 5 are failures, ¯\_(ツ)_/¯
+    // If we have no nodes left because all are failures, ¯\_(ツ)_/¯
```
Evalir:
```suggestion
    // If we have no nodes left it's because all are failures, ¯\_(ツ)_/¯
```

src/services/cherry-picker.ts
```diff
@@ -170,15 +170,14 @@ export class CherryPicker {
...
-    applicationID: string,
     serviceNode: string,
     elapsedTime: number,
     result: number,
-    timeout?: number,
-    pocketSession?: Session
+    session: Session,
+    timeout?: number
   ): Promise<void> {
     // Removed while load balancer cherry picking is off
```
Evalir, line 179: Is this comment still accurate?

src/services/cherry-picker.ts
```diff
@@ -216,7 +215,12 @@ export class CherryPicker {
     const serviceLog = await this.fetchRawServiceLog(blockchain, id)
 
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
Evalir: No need for this eslint disable anymore

src/services/metrics-recorder.ts
```diff
@@ -170,16 +170,8 @@ export class MetricsRecorder {
...
-        await this.cherryPicker.updateServiceQuality(
-          blockchainID,
-          applicationID,
-          serviceNode,
-          elapsedTime,
-          result,
-          timeout,
-          session
-        )
+      if (serviceNode && !Object.values(CheckMethods).includes(method as CheckMethods)) {
```
Evalir, line 173: Q: What does the second condition mean?

Evalir (approved):
Looks pretty good overall, just had a few questions when reading the code:

src/services/cherry-picker.ts
rem1niscence: nope, is an optional env to log cherry picker behavior

src/services/metrics-recorder.ts
rem1niscence: is not a value contained on that enum, or in practical terms is not a method of sync/chain check

src/services/cherry-picker.ts
rem1niscence: cc @nymd

PR #845, opened 2022-05-12
fix: sync check default vars

crisog:
If altruist request failed (catch error) the sync check would fail. This PR fixes this bug.


crisog:
I still need to fix the test case that should have covered this bug.

rem1niscence (approved):

PR #848, opened 2022-05-13
fix: parse json rpc err

crisog:
Pocket core responses to the API were returning invalid JSON-RPC, that made our parsing fail and mark some connection refused errors as user errors (not the case, these should be actual relay errors).

This PR fixes this issue and does proper parsing & handling on invalid JSON-RPC.

### DIFF
--- a/src/utils/errors.ts
+++ b/src/utils/errors.ts
@@ -5,6 +5,8 @@ export function isEVMError(response: string): boolean {
   try {
     const { code } = parseJSONRPCError(response)
 
+    // 3 is execution error
+    // -32000 itself can be thrown for what are server errors, all other -32000 are user errors
     return code === 3 || code < -32000
   } catch {
     return false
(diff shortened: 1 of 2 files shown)


src/utils/errors.ts
```diff
@@ -1,22 +1,11 @@
...
 export function isEVMError(response: string): boolean {
-  return isError(response, EVM_ERROR_CODES)
-}
-
-function isError(response: string, errorsToCheck: string[]): boolean {
   try {
     const { code } = parseJSONRPCError(response)
 
-    return errorsToCheck.some((error) => String(code).includes(error))
+    return code === 3 || code < -32000
```
blockjoe, line 10: maybe for clarity in the future since this isn't documented in the JSON RPC standard, something like

```js
if (code === 3) { // execution error
   return true
} elif (code < -32000) { // -32000 itself can be thrown for what are server errors, all other -32000 are user errors 
    return true
} return false
```

Or just some way of making it known why -32000 gets an excpetion

rem1niscence (approved):

PR #851, opened 2022-05-16
Remove-session-on-invalid-blockheight

rem1niscence:
Removes the session from cache when a invalid block height error is found


rem1niscence:
bad base branch

PR #857, opened 2022-05-26
Log-cherry-picker-env

rem1niscence:
Add infra config to log cherry picker stats as env


Evalir (approved):
gud code ser

rem1niscence:
> gud code ser

always glad to see people with good taste

PR #861, opened 2022-05-30
fix: enforce fallback response integrity & evm user error matching

crisog:
The fallback responses are the ones coming from our backup system. They come into play when the node selected from the network can't do it's job, and doesn't return any response. In our mission to be as fast as we can, we use our internal system to serve that request (doesn't generate revenue).

These responses were only enforced to a certain format, but the truth is that there are other several server errors from blockchain's node clients that indicate that the call wasn't effective. This PR adds some additional text enforcement to abstract these errors from users and let it fail without any specific blockchain's node message that could add confusion to end user.

Example error:
```
{"error":{"code":-32000,"message":"rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing dial tcp 127.0.0.1:9090: connect: connection refused\""},"id":732830328053361800,"jsonrpc":"2.0"}
```

### DIFF
--- a/src/services/pocket-relayer.ts
+++ b/src/services/pocket-relayer.ts
@@ -462,6 +462,12 @@ export class PocketRelayer {
             responseParsed = JSON.parse(fallbackResponse.data)
           }
 
+          const stringifiedResponse = JSON.stringify(responseParsed)
+
+          if (isRelayError(stringifiedResponse)) {
+            throw new Error(`Response is not valid: ${stringifiedResponse}`)
+          }
+
           this.metricsRecorder
             .recordMetric({
               requestID,
--- a/src/services/pocket-relayer.ts
+++ b/src/services/pocket-relayer.ts
@@ -464,7 +464,7 @@ export class PocketRelayer {
 
           const stringifiedResponse = JSON.stringify(responseParsed)
 
-          if (isRelayError(stringifiedResponse)) {
+          if (isRelayError(stringifiedResponse) && !isUserError(stringifiedResponse)) {
             throw new Error(`Response is not valid: ${stringifiedResponse}`)
           }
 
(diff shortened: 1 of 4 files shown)


crisog:
Another fix was also introduced at this PR, since it's error-related. It's the fix to the problem outlined in #862.

The current approach is matching server errors, and user error is anything with -32k error and NOT a server error. Since server errors are fewer, it is easier to handle and node runners won't be penalized by false positives on EVM unknown user error messages.

crisog:
Another fix was also introduced at this PR, since it's error-related. It's the fix to the problem outlined in #862.

The current approach is matching server errors, and user error is anything with -32k error and NOT a server error. Since server errors are fewer, it is easier to handle and node runners won't be penalized by false positives on EVM unknown user error messages.

src/services/pocket-relayer.ts
```diff
@@ -462,6 +462,12 @@ export class PocketRelayer {
             responseParsed = JSON.parse(fallbackResponse.data)
           }
 
+          const stringifiedResponse = JSON.stringify(responseParsed)
+
+          if (isRelayError(stringifiedResponse) && !isUserError(stringifiedResponse)) {
+            throw new Error(`Response is not valid: ${stringifiedResponse}`)
+          }
```
crisog, line 469: makes sure that the response obtained from altruist is a good one

src/utils/errors.ts
```diff
@@ -1,19 +1,19 @@
...
     const { code, message } = parseJSONRPCError(response)
 
-    const userError = EVM_USER_ERROR_MSGS.some((err) => message.includes(err))
+    const serverError = EVM_SERVER_ERROR_MSGS.some((err) => message.includes(err))
 
     // 3 is execution error
     // -32000 itself can be thrown for what are server errors
     // all other -32000 are user errors
-    return userError || code === 3 || code < -32000
+    return (!serverError && code === -32000) || code === 3 || code < -32000
```
crisog, line 16: anything thats not a server error & has error code -32k is assumed to be an user error, not node error.

rem1niscence (approved):

PR #882, opened 2022-06-23
Removes commit hash

rem1niscence:
Removes the commit hash prefix from the build by setting it blank as it might be reactivated later


rem1niscence:
Wrong branch

PR #887, opened 2022-06-23
revert: remove redis cache prefix completely

rem1niscence:
Completely removes redis prefix key


crisog (approved):
This removes the options from the library as it's supposed to be. Let's roll this out to canary but it is good to go.

PR #891, opened 2022-07-06
feat: add block height facets to sync behind log

crisog:
As of now, the block heights are only recorded in the error string. This is not very helpful to use in Datadog and we'll need it to make an strategy to balance out the sync allowance depending on how often nodes are out of sync.


adshmh:
Thanks for the PR. Does fisherman need updating with the same change?

PR #896, opened 2022-07-15
Add pub key metadata

rem1niscence:
Note: This does not modify the algorithm in any way
Add Application's public key to cherry picker metadata. This is so the cherry picker database can easily identified to which apps the active session belongs


rem1niscence:
Wrong starting branch

PR #900, opened 2022-07-21
Bring in fixes suggested by discord user msa6867#1823

blockjoe:
Given the user's comment about not being comfortable with git, wanted to just pull in the changes since they were quick so that the user could say if this is what they were suggesting.


crisog:
Cherry picked your commit @blockjoe to create a PR from a clean branch. This is the new PR #901.

PR #905, opened 2022-08-09
Chains mapping endpoint

rem1niscence:
Adds a new endpoint `blockchains/ids` where all the available chains are listed in a JSON object exposing their ID and API aliases


rem1niscence:
wrong base branch

PR #908, opened 2022-08-11
T-10526 - PHD Client Integration

commoddity:
Creating this PR to create functionality to pull Portal data from PHD (Pocket HTTP Database) and fall back to MongoDB if fails.

Main functionality is written but will need extensive testing.


src/controllers/blockchains.controller.ts
```diff
@@ -75,15 +94,21 @@ export class BlockchainsController {
...
 
     const aliases = {}
-    blockchains.forEach((blockchain) => {
-      aliases[blockchain.description] = {
-        id: blockchain.hash,
-        prefix: blockchain.blockchainAliases,
+
+    blockchains.forEach(({ id, blockchainAliases, description }) => {
+      aliases[description] = {
+        id: id,
```
crisog:
```suggestion
        id,
```

src/services/phd-client.ts
```diff
@@ -0,0 +1,177 @@
...
+  }
+
+  async find<T extends Entity>({
+    path,
+    model,
+    cache,
+    cacheKey = 'blockchains', // Currently .find only used to get blockchains
+    fallback,
+  }: FindParams<T>): Promise<T[]> {
+    const url = `${this.baseUrl}/${path}`
```
crisog, line 75: What do you think about creating some sort of Enum for the routes/path of PHD? 

Example:
https://github.com/pokt-foundation/pocket-js/blob/8e71ab6caccd1aa43ee3b8532e791165ed30f471/packages/abstract-provider/src/routes.ts#L11-L17

Ideally, we want some sort of versioning on PHD as we have in the API (f.g. /v1) so it is easier to upgrade later.

This is just an enhancement, won't block the PR because of this. Let me know what you think!

tests/acceptance/blockchain.controller.acceptance.tests.ts
```diff
@@ -24,129 +29,178 @@ describe('Blockchains controller (acceptance)', () => {
...
-    expect(res.body).to.be.Array()
-    expect(res.body).to.have.length(1)
-    expect(res.body[0]).to.containEql(expected)
+  afterEach(async () => {
+    await cleanDB()
+    axiosMock.reset()
+  })
+
+  describe('/blockchains/count endpoint', () => {
+    it('retrieves the total count of blockchains from  from the Pocket HTTP DB', async () => {
```
crisog:
```suggestion
    it('retrieves the total count of blockchains from the Pocket HTTP DB', async () => {
```

crisog (approved):
Looks good to me. Nice job on this 👍🏽

src/services/phd-client.ts
commoddity: I think this is a good idea but it requires some fiddling with Typescript that I don't think is worth the lift given everything else going on. Going to suggest that if we do this we wait until after go-live.

tests-integration/phd-client.integration.ts
```diff
@@ -0,0 +1,215 @@
+import { expect, sinon } from '@loopback/testlab'
```
rem1niscence: this is just a nit-pick but in loopback terms, acceptance tests == integration tests. So you could move this file and/or folder within the acceptance folder so it goes more like the loopback's recommended folder structure

commoddity: Makes sense. Actually the only reason I put these tests there was to have an easier time running them on their own but I have no objection to moving them wherever you feel is best.

src/application.ts
```diff
@@ -144,6 +149,11 @@ export class PocketGatewayApplication extends BootMixin(ServiceMixin(RepositoryM
 
     this.bind('cache').to(cache)
 
+    // Bind PHD Client
```
rem1niscence, line 160: before doing the bind, a health check should be done as this an external service, you can check #944 in the same `src/applications.ts` so you can have an example on how it can be done

src/controllers/blockchains.controller.ts
```diff
@@ -57,7 +69,12 @@ export class BlockchainsController {
     @param.filter(Blockchains, { exclude: 'where' })
     filter?: FilterExcludingWhere<Blockchains>
   ): Promise<Blockchains> {
-    return this.blockchainsRepository.findById(id, filter)
+    return this.phdClient.findById({
+      path: 'blockchain',
```
rem1niscence: here (and everywhere else that you're specifying the path) should be better done with an ENUM of paths in order to avoid typos and change it from one single place if needed

src/controllers/v1.controller.ts
```diff
@@ -667,10 +677,13 @@ export class V1Controller {
 
     if (!cachedLoadBalancer) {
       try {
-        const loadBalancer = await this.loadBalancersRepository.findById(id, filter)
-
-        await this.cache.set(id, JSON.stringify(loadBalancer), 'EX', 60)
-        return new LoadBalancers(loadBalancer)
+        return await this.phdClient.findById({
+          path: 'load_balancer',
```
rem1niscence: same here, avoid raw strings, better use an enum

src/controllers/v1.controller.ts
```diff
@@ -684,10 +697,13 @@ export class V1Controller {
 
     if (!cachedApplication) {
       try {
-        const application = await this.applicationsRepository.findById(id, filter)
-
-        await this.cache.set(id, JSON.stringify(application), 'EX', 60)
-        return new Applications(application)
+        return await this.phdClient.findById({
+          path: 'application',
```
rem1niscence: same here

src/models/load-balancers.model.ts
```diff
@@ -60,14 +60,20 @@ export class LoadBalancers extends Entity {
     type: 'string',
     required: true,
   })
-  user: string
+  userID: string
```
rem1niscence: which schema is the "main" schema here? mongo or postgres? as `userID` exist in postgres not mongo, but if the main schema to follow now is postgres then all good

src/services/phd-client.ts
```diff
@@ -0,0 +1,200 @@
+import axios from 'axios'
+import 'dotenv/config'
```
rem1niscence, line 2: why this import?

tests-integration/phd-client.integration.ts
rem1niscence: all good, going with the recommended way then is the one I mentioned above

rem1niscence (dismissed):
left a few comments but none of them are blocking, although recommend to check on them

src/controllers/blockchains.controller.ts
commoddity: Sure, I can add that.

src/models/load-balancers.model.ts
commoddity: There are some newer apps that were created during development that have `userID` in MongoDB so I wasn't sure if we should include both. 

In the MongoDB update we made to PUB I set it to add both fields on new apps to be safe. But I think all of those apps were just test apps we made during development so maybe we don't need to care about `userID` here?

src/services/phd-client.ts
commoddity: Was for the integration tests to work.

src/application.ts
commoddity: OK, added a health check in `application.ts`

src/controllers/blockchains.controller.ts
commoddity: Done. All paths are now an enum.

tests-integration/phd-client.integration.ts
commoddity: Done. PHD Client integration tests moved to `tests/acceptance`

src/controllers/v1.controller.ts
commoddity: Done. All paths are now an enum.

src/controllers/v1.controller.ts
commoddity: Done. All paths are now an enum.

src/models/load-balancers.model.ts
commoddity: Updated to use `user` and made a change to the PHD client to change `userID` to `user` when fetching for PHD (if `user` exists)

src/services/phd-client.ts
commoddity: Was able to remove it with the new test changes.

src/services/phd-client.ts
```diff
@@ -0,0 +1,213 @@
...
+    this.apiKey = apiKey
+  }
+
+  async find<T extends Entity>({ path, model, cache, cacheKey, fallback }: FindParams<T>): Promise<T[]> {
+    if (cache && !cacheKey) {
+      throw new Error(`cacheKey not set for path ${path}`)
+    }
+
+    const url = `${this.baseUrl}/${path}`
+    const modelFields = this.getRequiredModelFields(model)
```
crisog, line 76: I think this should be inside the try catch since there isn't any on `getRequiredModelFields`

src/services/phd-client.ts
```diff
@@ -0,0 +1,213 @@
...
+    if (cache && cacheKey) {
+      await cache.set(cacheKey, JSON.stringify(modelsData), 'EX', 60)
+    }
+
+    return modelsData
+  }
+
+  async findById<T extends Entity>({ path, id, model, cache, fallback }: FindOneParams<T>): Promise<T> {
+    const url = `${this.baseUrl}/${path}/${id}`
+    const modelFields = this.getRequiredModelFields(model)
```
crisog, line 116: Same here

crisog (changes requested):

src/services/phd-client.ts
commoddity: I don't think `getRequiredModelFields` can throw an error though? It's not async and AFAIK there's not really any way for it to fail.

commoddity: The try/catch in these methods is specifically meant to capture errors in fetching data from PHD and falling back to getting them from MongoDB if necessary.

src/controllers/blockchains.controller.ts
```diff
@@ -1,11 +1,15 @@
+import { inject } from '@loopback/context'
 import { Count, CountSchema, Filter, FilterExcludingWhere, repository, Where } from '@loopback/repository'
 import { param, get, getModelSchemaRef } from '@loopback/rest'
+
```
rem1niscence: remove this line

tests/acceptance/blockchain.controller.acceptance.tests.ts
```diff
@@ -1,6 +1,9 @@
+import axios from 'axios'
+import MockAdapter from 'axios-mock-adapter'
 import { Client, expect, sinon } from '@loopback/testlab'
+
```
rem1niscence: remove this line

rem1niscence (dismissed):

tests/acceptance/blockchain.controller.acceptance.tests.ts
commoddity: Done

rem1niscence (approved):

crisog (approved):

adshmh (approved):

PR #926, opened 2022-08-13
[staging] fix: use local caching for blocked addresses

crisog:
This PR also includes a refinement to the logic blocking addresses from a list.


adshmh (approved):
LGTM

PR #931, opened 2022-08-17
fix: empty params object error

rem1niscence:
Fixes the error when trying to decode the address of a contract which may not specify it,. Defaults to undefined as the default case of the switch where is being consumed


src/utils/evm/parsing.ts
```diff
@@ -3,7 +3,7 @@ import { parseMethod } from '../parsing'
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export function extractContractAddress(rawData: Record<string, any>): { fromAddress: string; toAddress: string } {
-  const params = rawData.params
+  const params = rawData.params !== undefined ? rawData.params : {}
```
adshmh, line 6: Could we add a unit test to cover this?

adshmh (changes requested):

PR #932, opened 2022-08-28
First integration with rate-limiter service: log warning only

adshmh:
Signed-off-by: Arash Deshmeh <adshmh@gmail.com>

### DIFF
--- a/src/controllers/v1.controller.ts
+++ b/src/controllers/v1.controller.ts
@@ -15,7 +15,7 @@ import { CherryPicker } from '../services/cherry-picker'
 import { MetricsRecorder } from '../services/metrics-recorder'
 import { PocketRelayer } from '../services/pocket-relayer'
 import { SyncChecker } from '../services/sync-checker'
-import { checkWhitelist } from '../utils/enforcements'
+import { checkWhitelist, shouldRateLimit } from '../utils/enforcements'
 import { parseRawData, parseRPCID } from '../utils/parsing'
 import { getBlockchainAliasesByDomain, loadBlockchain } from '../utils/relayer'
 import { SendRelayOptions } from '../utils/types'
@@ -65,6 +65,7 @@ export class V1Controller {
     @inject('archivalChains') private archivalChains: string[],
     @inject('alwaysRedirectToAltruists') private alwaysRedirectToAltruists: boolean,
     @inject('dispatchURL') private dispatchURL: string,
+    @inject('rateLimiterURL') private rateLimiterURL: string,
     @repository(ApplicationsRepository)
     public applicationsRepository: ApplicationsRepository,
     @repository(BlockchainsRepository)
@@ -312,6 +313,23 @@ export class V1Controller {
         throw new ErrorObject(reqRPCID, new jsonrpc.JsonRpcError('No application found in the load balancer', -32055))
       }
 
+      if (!gigastakeOptions.gigastaked) {
+        const shouldLimit = await shouldRateLimit(application.id, this.rateLimiterURL, this.cache)
+        if (shouldLimit) {
+          logger.log(
+            'warn',
+            'relay count on application associated with the endpoint has exceeded the rate limit ' + application.id,
+            {
+              requestID: this.requestID,
+              relayType: 'LB',
+              typeID: id,
+              serviceNode: '',
+              origin: this.origin,
+            }
+          )
+        }
+      }
+
       if (gigastakeOptions?.gatewaySettings) {
         application.gatewaySettings = gigastakeOptions.gatewaySettings
       }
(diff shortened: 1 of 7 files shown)


adshmh:
Link T-12166

src/controllers/v1.controller.ts
```diff
@@ -312,6 +312,26 @@ export class V1Controller {
         throw new ErrorObject(reqRPCID, new jsonrpc.JsonRpcError('No application found in the load balancer', -32055))
       }
 
+      if (!gigastakeOptions.gigastaked) {
+        const { shouldLimit, numRateLimitedApps } = await shouldRateLimit(application.id, this.cache)
+        if (numRateLimitedApps === 0) {
+          logger.log('warn', 'Rate-limited applications list is empty; rate-limiting disabled')
```
rem1niscence: this warn log would appear on each relay in case the list is empty, which can occur large log ingestion charges from our log management system and is something that, from a coding side, we already know if we left the list empty. I recommend to perform the check of existing apps in rate limit in `applications.ts` and throw the warning there so is made only once

src/controllers/v1.controller.ts
```diff
@@ -428,6 +448,20 @@ export class V1Controller {
       const applicationID = application.id
       const applicationPublicKey = application.gatewayAAT.applicationPublicKey
 
+      const { shouldLimit, numRateLimitedApps } = await shouldRateLimit(applicationID, this.cache)
+      if (numRateLimitedApps === 0) {
+        logger.log('warn', 'Rate-limited applications list is empty; rate-limiting disabled')
```
rem1niscence: same comment as above

src/observers/environment.observer.ts
```diff
@@ -37,6 +37,7 @@ export class EnvironmentObserver implements LifeCycleObserver {
     'INFLUX_ORG',
     'ARCHIVAL_CHAINS',
     'BLOCKED_ADDRESSES_URL',
+    'RATE_LIMITER_URL',
```
rem1niscence: is this truly required? should rate limiter be on by default or should it be a toggle?

src/utils/cache.ts
```diff
@@ -107,3 +107,42 @@ export async function getBlockedAddresses(redis: Redis, URL: string): Promise<st
...
+  if (!cachedRateLimitedApps) {
+    try {
+      const axiosConfig = {
+        method: 'GET',
+        url: URL,
+        timeout: 10000,
+      } as AxiosRequestConfig
+
+      const { data } = await axios(axiosConfig)
+      const { ApplicationIDs: rateLimitedAppsList } = data
```
rem1niscence:
```suggestion
      const { applicationIDs: rateLimitedAppsList } = data
```

src/utils/enforcements.ts
```diff
@@ -17,6 +19,33 @@ export function checkEnforcementJSON(test: string): boolean {
...
+  let obj = {
+    shouldLimit: false,
+    numRateLimitedApps: 0,
+  }
+
+  if (appID.length === 0) {
+    return obj
+  }
+
+  const url = process.env.RATE_LIMITER_URL ?? ''
```
rem1niscence, line 28: the portal-api usually reads the url as a global variable in top of the file or is passed as a function, I personally prefer the later as is more clear where the value is coming from but I recommend one of the two for consistency

rem1niscence (changes requested):

src/observers/environment.observer.ts
adshmh: This allows:

1. Configuration of new endpoint for `rate-limiter` if necessary
2. Toggling `rate-limiter` on/off if any potentially urgent issues arise.

src/utils/cache.ts
adshmh: This has to stay upper-case to match the output of `rate-limiter`(otherwise we will miss the output)

rem1niscence: All right

src/observers/environment.observer.ts
rem1niscence: Not providing the env there will cause an error on the api (as envs declared there must be provided) so I don't see how can it be toggle from there. There's another array on that same file of optional envs, you could add it to that array and check for that in `applications.ts` as is the file where services are boostraped and check on api boot

src/controllers/v1.controller.ts
adshmh: Thank you for the review. Fixed.

src/controllers/v1.controller.ts
adshmh: Thank you for the review. Fixed.

src/observers/environment.observer.ts
adshmh: Thank you for the review. Fixed.

src/utils/enforcements.ts
adshmh: Thank you for the review. Fixed.

rem1niscence:
Lgtm

rem1niscence (approved):

kutoft:
Close T-12166

PR #952, opened 2022-09-16
Use a larger-than-default batch size for writes to influx

adshmh:
Use a larger-than-default batch size for influxdb writes


rem1niscence (approved):

crisog (approved):
Theres an unused variable `WriteOptions` that can be removed but other than that it looks good!

adshmh:
> Theres an unused variable `WriteOptions` that can be removed but other than that it looks good!

Thanks, fixed. Please take another look.

PR #958, opened 2022-09-20
feat: Rate Limiter token

crisog:
Adds the token to the rate limiter


.github/workflows/production-ap-southeast-2.yml
```diff
@@ -56,13 +56,15 @@ jobs:
           envkey_LOG_TO_CLOUDWATCH: false
           envkey_AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
           envkey_AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          envkey_COMMIT_HASH: ${{ github.sha }}
```
adshmh, line 59: is this change intentional?

crisog: These files are auto generated, it was missing in that one file but it is present in all of the other regions. I think it is fine.

adshmh (approved):
LGTM

PR #964, opened 2022-10-04
Multiple influx instances

rem1niscence:
Adds the ability to write to multiple influx instances at the same time by only updating an ENV


kutoft:
Link T-14608

kutoft:
Close T-14608

PR #965, opened 2022-10-06
feat: add region to relay headers

crisog:
This PR adds a header containing the region to relays (including check relays) - this will greatly help the community to play around with their Geo-locations and see clearly where the traffic is coming from.

### DIFF
--- a/src/application.ts
+++ b/src/application.ts
@@ -69,6 +69,7 @@ export class PocketGatewayApplication extends BootMixin(ServiceMixin(RepositoryM
       REDIS_LOCAL_TTL_FACTOR,
       RATE_LIMITER_URL,
       RATE_LIMITER_TOKEN,
+      REGION,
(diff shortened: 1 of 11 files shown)


crisog:
T-14751

crisog:
Link T-14751

nodiesBlade:
Hey @rem1niscence and @adshmh 

When is this expected to be merged/tested? It would greatly help with us [with our monitoring tool](https://github.com/baaspoolsllc/nodies_monitoring/tree/staging). We aggregate all incoming log traffic to provide real time analysis. Currently we only have a handle of remote ip and would prefer not to resort to any geo-ip mapping using a licensed database. If we are able to retrieve region information from headers, it could help the pokt community greatly.

- Better observability for entire pocket networks (including latency and regional live traffic)
- Node runners can be less dependent of datadog traffic

PR #968, opened 2022-10-10
[master] Add multiple influx instances

rem1niscence:
merges #964


adshmh (commented):
Thank you for the PR. 
This makes multiple significant changes to how we record metrics.
I think acceptance tests are necessary to ensure the following:
 - ALL the specified instances get the metrics
 - A failure (or long delay) when writing to 1 instance doesn't affect others
 - A long delay when writing to any of the instances does not affect relay latency.
 - Any Error can be traced to the corresponding instance.

PR #969, opened 2022-10-12
Public RPC POC

kutoft:
Link T-14836

### DIFF
--- a/src/controllers/v1.controller.ts
+++ b/src/controllers/v1.controller.ts
@@ -77,6 +77,7 @@ export class V1Controller {
     @inject('dispatchURL') private dispatchURL: string,
     @inject('rateLimiterURL') private rateLimiterURL: string,
     @inject('rateLimiterToken') private rateLimiterToken: string,
+    @inject('publicRPCEthAddress') private publicRPCEthAddress: string,
     @repository(ApplicationsRepository)
     public applicationsRepository: ApplicationsRepository,
     @repository(BlockchainsRepository)
@@ -242,7 +243,7 @@ export class V1Controller {
 
       // MEV: Where should this actually live???
       if (blockchainRequest.toLowerCase() === 'eth-rpc' && isEthSendMethod(parsedRawData)) {
-        const BEVResponse = await handleBEVRequest(this.requestID, parsedRawData)
+        const BEVResponse = await handleBEVRequest(this.requestID, parsedRawData, this.publicRPCEthAddress)
         if (BEVResponse) {
           return JSON.stringify({ jsonrpc: '2.0', id: reqRPCID, result: BEVResponse })
         }
--- a/src/controllers/v1.controller.ts
+++ b/src/controllers/v1.controller.ts
@@ -1,5 +1,6 @@
 import { Relayer } from '@pokt-foundation/pocketjs-relayer'
 import { HTTPMethod } from '@pokt-foundation/pocketjs-types'
+import { ethers } from 'ethers'
 import jsonrpc, { ErrorObject, JsonRpcError } from 'jsonrpc-lite'
 import { Pool as PGPool } from 'pg'
 import { inject } from '@loopback/context'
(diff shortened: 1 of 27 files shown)


src/utils/mev.ts
```diff
@@ -0,0 +1,77 @@
+import axios from 'axios'
+import { ethers } from 'ethers'
+const logger = require('../services/logger')
+
+const bevUrl = 'https://bev-relay.kolibr.io'
+
+export function isEthSendMethod(rawData) {
+  if (rawData instanceof Array) {
```
rem1niscence, line 8: why this is false on an array? there's a possibility that one or more of the methods within the array is `eth_sendRawTransaction`

src/utils/mev.ts
```diff
@@ -0,0 +1,77 @@
...
+    console.log(e.message)
+  }
+
+  return false
+}
+
+export async function handleBEVRequest(id, parsedRawData, address) {
+  // store address in Github secrets
+  const checkBevResult = await checkBevRequest(id, parsedRawData, address)
+  logger.log('Check BEV result:', checkBevResult)
```
rem1niscence, line 61: the first argument of `logger.log` must be the type of log (as in `debug, infro, warn, errror`)

src/utils/mev.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  if (!checkBevResult) {
+    logger.log('Request error')
```
rem1niscence, line 64: same here

src/utils/mev.ts
```diff
@@ -0,0 +1,77 @@
...
+      id: requestId,
+      method: 'submit_bev',
+      params: {
+        tx_raw_data: txRawData,
+        bundle_hash: bundleHash,
+      },
+    })
+    return res?.data?.result
+  } catch (e) {
+    console.log(e.message)
```
rem1niscence, line 52: better use `logger.log` instead of `console.log`

src/utils/mev.ts
```diff
@@ -0,0 +1,77 @@
+import axios from 'axios'
+import { ethers } from 'ethers'
+const logger = require('../services/logger')
+
+const bevUrl = 'https://bev-relay.kolibr.io'
```
rem1niscence, line 5: this should be on a environment variable IMO, not only for privacy but to quickly change the URL if needed

src/utils/mev.ts
rem1niscence: also in the function that this is being called we are already logging the error with info, here we are not logging any info, this could be removed to avoid two logs

rem1niscence (changes requested):

src/utils/mev.ts
```diff
@@ -0,0 +1,77 @@
...
+  }
+  return false
+}
+
+export async function checkBevRequest(requestId, txRawData, broadcasterAddress) {
+  try {
+    const res = await axios.post(bevUrl, {
+      jsonrpc: '2.0',
+      id: requestId,
+      method: 'check_bev',
```
apadenko, line 22: `check_bev` requires an authentication header in order to eliminate continuous spam checks without submission. 

`check_and_submit_bev` doesn't require authentication. If BEV exists - it will automatically submit the bundle to builders. 

So you should either store secret and use it for `check_bev` authorization 
or use public `check_and_submit_bev`

src/utils/mev.ts
apadenko, line 5: HTTP is good for POC and testing. 
But WS is more recommended, it will dramatically reduce latency.

src/controllers/v1.controller.ts
```diff
@@ -230,6 +233,16 @@ export class V1Controller {
 
       reqRPCID = parseRPCID(parsedRawData)
 
+      const [blockchainRequest] = this.host.split('.')
+
+      // MEV: Where should this actually live???
+      if (blockchainRequest.toLowerCase() === 'eth-rpc' && isEthSendMethod(parsedRawData)) {
+        const BEVResponse = await handleBEVRequest(this.requestID, parsedRawData, this.publicRPCEthAddress)
+        if (BEVResponse) {
```
apadenko, line 241: IF BEVResponse is successful you should still perform existing broadcast logic, 
but with 1 block (10-12sec) delay. being not dependent on the BEV relay.

src/utils/mev.ts
blockjoe: It would be pretty difficult to split off an array and inject the correct response back into the array of the transactions in the order it came through that the consumer would expect.

PR #970, opened 2022-10-13
Only altruist chains

rem1niscence:
Adds the ability to only set X amount of only altruist chains based on an environment variable


src/services/cherry-picker.ts
```diff
@@ -266,7 +266,7 @@ export class CherryPicker {
           p90: bucketedServiceQuality.p90,
           attempts: unsortedLog.attempts,
           successRate: unsortedLog.successRate,
-          applicationPublicKey: session.header.applicationPubKey,
+          applicationPublicKey: session?.header?.applicationPubKey,
```
adshmh: Please make this change separately, ideally as a separate PR.

src/services/cherry-picker.ts
```diff
@@ -283,13 +283,13 @@ export class CherryPicker {
         results: results,
         medianSuccessLatency: elapsedTime.toFixed(5),
         weightedSuccessLatency: elapsedTime.toFixed(5),
-        sessionKey: session.key,
-        sessionHeight: session.header.sessionBlockHeight,
+        sessionKey: session?.key,
```
kutoft: what changed that these are now needed?

src/services/metrics-recorder.ts
```diff
@@ -129,8 +129,8 @@ export class MetricsRecorder {
           blockchainID,
           sessionKey,
           sticky,
-          sessionBlockHeight: session.header.sessionBlockHeight,
-          blockHeight: session.blockHeight,
+          sessionBlockHeight: session?.header.sessionBlockHeight,
```
adshmh, line 134: Same with these changes: they do not seem relevant to the topic of this PR, so they are best made through a separate PR (with accompanying unit tests)

src/services/metrics-recorder.ts
```diff
@@ -129,8 +129,8 @@ export class MetricsRecorder {
...
+          blockHeight: session?.blockHeight,
```
adshmh, line 135: Please see previous comments on unrelated changes.

src/services/pocket-relayer.ts
```diff
@@ -227,7 +231,7 @@ export class PocketRelayer {
     const fallbackAvailable = blockchainAltruist ? true : false
 
     try {
-      if (!this.alwaysRedirectToAltruists) {
+      if (!this.altruistOnlyChains.includes(blockchainID) && !this.alwaysRedirectToAltruists) {
```
adshmh: Could we add an acceptance test to cover this please?

Also, why is `!this.` added to `alwaysRedirectToAltruists` check? Was it not working as expected already?

adshmh (changes requested):
Thank you for the PR. Just a few changes needed. Please resend for review again once the review comments are addressed.

src/services/cherry-picker.ts
rem1niscence: those changes are needed, otherwise the api will break

src/services/metrics-recorder.ts
rem1niscence: they all are, those are for request without those values which will crash as they will come as undefined

src/services/pocket-relayer.ts
rem1niscence: because this is a second path after the if, the conditions of the if are to not enter it so we are negating the situation, (as in, if this is value is there, then skip the if)

adshmh: I see, then at-least 1 test case to cover the new scenario, i.e. blockchain in the list of altruists-only chains.

tests/acceptance/v1.controller.acceptance.ts
```diff
@@ -1982,5 +1982,38 @@ describe('V1 controller (acceptance)', () => {
       expect(response.body).to.have.property('error')
       expect(response.body.error.message).to.startWith('Restricted endpoint: contract address not allowed')
     })
+
+    it('invokes GET /v1/{appId} and successfully relays a request only through the altruist', async () => {
```
adshmh: Thank you for adding the test.
This should be `invokes POST`

tests/acceptance/v1.controller.acceptance.ts
```diff
@@ -1982,5 +1982,38 @@ describe('V1 controller (acceptance)', () => {
...
+        .set('host', 'eth-mainnet-x')
+        .expect(200)
+
+      expect(response.headers).to.containDeep({ 'content-type': 'application/json' })
+      expect(response.body).to.have.properties('id', 'jsonrpc', 'result')
+      expect(parseInt(response.body.result, 16)).to.be.aboveOrEqual(0)
+
+      const expectedAltruistLog = logSpy.calledWith(
+        'info',
+        sinon.match((arg: string) => arg.startsWith('SUCCESS FALLBACK RELAYING 0041'))
```
adshmh: Does this distinguish between altruist-only chains, and chains where the altruist is used as fallback?
If not, we likely need to log differently in the controller for altruist-only chains.

tests/acceptance/v1.controller.acceptance.ts
rem1niscence: it does, is a typo, fixed

tests/acceptance/v1.controller.acceptance.ts
rem1niscence: it doesn't, it can be added tho

tests/acceptance/v1.controller.acceptance.ts
```diff
@@ -1982,5 +1982,40 @@ describe('V1 controller (acceptance)', () => {
...
+      const response = await client
+        .post('/v1/sd9fj31d714kgos42e68f9gh')
+        .send({ method: 'eth_blockNumber', id: 1, jsonrpc: '2.0' })
+        .set('Accept', 'application/json')
+        .set('host', 'eth-mainnet-x')
+        .expect(200)
+
+      expect(response.headers).to.containDeep({ 'content-type': 'application/json' })
+      expect(response.body).to.have.properties('id', 'jsonrpc', 'result')
+      expect(parseInt(response.body.result, 16)).to.be.aboveOrEqual(0)
```
kutoft, line 2002: ignorant question here. Ive seen this in like all our tests. But what exactly is it telling us? Feels pretty flimsy.

rem1niscence: is an example response from a node, that is the general format but the values inside doesn't really matter as long as they are the ones to be expected

src/services/pocket-relayer.ts
rem1niscence: done

kutoft (approved):
LGTM!

adshmh (approved):
LGTM. Thanks!

PR #975, opened 2022-10-18
feat: add blockchain subdomain facet

crisog:
Link T-14959

PR #977, opened 2022-10-18
fix: remove domain overwrite on simple redirects

crisog:
This is a complementary PR for #975. 

The way that simple redirects were designed does not allow for a nice way to log the redirect domain, because it gets overwritten for the specific blockchain alias. This PR removes this domain overwrite to be able to log the redirect domain correctly.

### DIFF
--- a/src/controllers/v1.controller.ts
+++ b/src/controllers/v1.controller.ts
@@ -18,7 +18,7 @@ import { PocketRelayer } from '../services/pocket-relayer'
 import { SyncChecker } from '../services/sync-checker'
 import { checkWhitelist, RateLimiter, shouldRateLimit } from '../utils/enforcements'
 import { parseRawData, parseRPCID } from '../utils/parsing'
-import { getBlockchainAliasesByDomain, loadBlockchain } from '../utils/relayer'
+import { loadBlockchain } from '../utils/relayer'
 import { SendRelayOptions } from '../utils/types'
 const logger = require('../services/logger')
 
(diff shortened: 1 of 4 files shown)


crisog:
Link T-14959

src/controllers/v1.controller.ts
```diff
@@ -273,7 +257,7 @@ export class V1Controller {
           if (!loadBalancer?.id) {
             throw new ErrorObject(
               reqRPCID,
-              new jsonrpc.JsonRpcError(`GS (${redirect.alias}) load balancer not found`, -32054)
+              new jsonrpc.JsonRpcError(`${redirect.alias} gigastake load balancer not found`, -32054)
```
adshmh, line 260: Please remove this change if it is not strictly required for the purpose of the PR.

src/controllers/v1.controller.ts
```diff
@@ -767,7 +751,10 @@ export class V1Controller {
     const loadBalancer = await this.fetchLoadBalancer(redirect.loadBalancerID, filter)
 
     if (!loadBalancer?.id) {
-      throw new ErrorObject(rpcID, new jsonrpc.JsonRpcError(`GS (${redirect.alias}) load balancer not found`, -32054))
+      throw new ErrorObject(
```
adshmh, line 754: Please remove this change if it is not strictly required for the purpose of the PR.

src/utils/relayer.ts
```diff
@@ -132,38 +142,3 @@ export async function loadBlockchain(
     blockchainRedirects,
   } as BlockchainDetails)
 }
-
-// Get blockchain's alias by it's redirect domain
-export async function getBlockchainAliasesByDomain(
```
adshmh, line 137: Is there a specific reason why this could not be used, as it is already performing the match on redirects.

adshmh (changes requested):
Thank you for the PR. A few comments to address.

kutoft:
Link T-15008

rem1niscence (approved):
LGTM

kutoft:
Link T-15008

crisog:
#984 ENV can be removed when this gets merged.

PR #980, opened 2022-10-19
[master] feat: add public rpc relay log

crisog:
PR #978


kutoft (approved):
LGTM

src/controllers/v1.controller.ts
```diff
@@ -140,6 +140,13 @@ export class V1Controller {
 
       rpcID = parseRPCID(parsedRawData)
 
+      const [blockchainRequest] = this.host.split('.')
+
+      logger.log('info', `PUBLIC RPC RELAY REDIRECTED FOR ${blockchainRequest}`, {
```
adshmh, line 145: Ideally this should be accompanied by a test for logger output. But that is not a blocker IMO.

adshmh (approved):
Thank you for the PR. Just 1 minor comment. Otherwise LGTM.

PR #984, opened 2022-11-14
Gateway host env

rem1niscence:
move gateway host string to an environment variable


tests/acceptance/v1.controller.acceptance.ts
```diff
@@ -1983,8 +1983,7 @@ describe('V1 controller (acceptance)', () => {
       expect(response.body.error.message).to.startWith('Restricted endpoint: contract address not allowed')
     })
 
-    // eslint-disable-next-line mocha/no-exclusive-tests
-    it.only('invokes POST /v1/{appId} and successfully relays a request only through the altruist', async () => {
+    it('invokes POST /v1/{appId} and successfully relays a request only through the altruist', async () => {
```
adshmh, line 1991: This doesn't related to this PR, could you clarify why it is needed?

adshmh (changes requested):
Could we add tests to cover both valid and invalid sub-domain sceanrios?

crisog (dismissed):
Looks good, straight forward 👍🏽 

If #977 gets merged, this ENV won't be needed in the future so I'll add a note in there.

adshmh (approved):
LGTM. Thanks!

PR #985, opened 2022-11-14
FEAT: Add loki support

Gustavobelfort:
This:
- Adds support to loki as a Winston log transport endpoint
- Adds required env vars to loki into the environment samples

### DIFF
--- a/src/services/logger.ts
+++ b/src/services/logger.ts
@@ -123,7 +123,7 @@ const options = {
     host: lokiHost,
     basicAuth: lokiBasicAuth,
     json: true,
-    labels: { app: 'portal-api' },
+    labels: { app: logName },
     format: format.json(),
     replaceTimestamp: true,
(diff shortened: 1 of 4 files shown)


src/services/logger.ts
```diff
@@ -114,6 +119,15 @@ const options = {
     intakeRegion: 'eu',
     ddtags: `region:${region}`,
   },
+  loki: {
+    host: lokiHost,
+    basicAuth: lokiBasicAuth,
+    json: true,
+    labels: { app: 'portal-api' },
```
crisog:
```suggestion
    labels: { app: logName },
```

crisog (changes requested):

src/services/logger.ts
Gustavobelfort: Done

crisog (approved):

PR #992, opened 2022-11-21
Log URL

rem1niscence:
log full url path as part of the result log response


adshmh (approved):
LGTM. Thanks!

src/services/pocket-relayer.ts
```diff
@@ -121,6 +125,7 @@ export class PocketRelayer {
     this.alwaysRedirectToAltruists = alwaysRedirectToAltruists
     this.altruistOnlyChains = altruistOnlyChains
     this.dispatchers = dispatchers
+    this.requestURL = `${request?.headers?.host}${request?.url}`
```
crisog, line 128: Not sure about this so I'll ask - doesn't URL include host already?

PR #994, opened 2022-11-25
ref: recorder optimization

crisog:
Logs volume is already too big. The shorter the logs, the cheaper the bill will be.

The end goal is to shift from logs to metrics as much as possible.


adshmh (approved):
LGTM

rem1niscence (approved):

PR #1006, opened 2022-12-05
feat: gnosis merge checker

crisog:
Preparing the merge checker to support Gnosis Merge.

### DIFF
--- a/src/services/merge-checker.ts
+++ b/src/services/merge-checker.ts
@@ -145,6 +145,7 @@ export class MergeChecker {
       }
 
       if (
+        blockchain &&
         BigInt(nodeTotalDifficulty) === TERMINAL_TOTAL_DIFFICULTY[blockchain] &&
         nodeBlockNumber >= MERGE_BLOCK_NUMBER[blockchain]
       ) {
--- a/src/services/merge-checker.ts
+++ b/src/services/merge-checker.ts
@@ -16,8 +16,15 @@ import { Cache } from './cache'
 
 const logger = require('../services/logger')
 
-const MERGE_BLOCK_NUMBER = 15537394
-const TERMINAL_TOTAL_DIFFICULTY = BigInt('58750003716598352816469')
+const MERGE_BLOCK_NUMBER = {
+  ethereum: 15537394,
+  gnosis: 15537394, // Needs to be updated to the proper block height
+}
+
+const TERMINAL_TOTAL_DIFFICULTY = {
+  ethereum: BigInt('58750003716598352816469'),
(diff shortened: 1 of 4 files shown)


src/services/merge-checker.ts
```diff
@@ -16,8 +16,15 @@ import { Cache } from './cache'
...
-const MERGE_BLOCK_NUMBER = 15537394
-const TERMINAL_TOTAL_DIFFICULTY = BigInt('58750003716598352816469')
+const MERGE_BLOCK_NUMBER = {
+  ethereum: 15537394,
+  gnosis: 15537394, // Needs to be updated to the proper block height
+}
+
+const TERMINAL_TOTAL_DIFFICULTY = {
+  ethereum: BigInt('58750003716598352816469'),
+  gnosis: BigInt('8626000000000000000000058750000000000000000000'),
```
blockjoe: This will have to be changed to whatever value the Total Difficulty is on the first block after it crosses over the 8626000000000000000000058750000000000000000000 value.

When this happened with the Eth merge, it meant us sitting up and watching for this value: 58750003716598352816469, as the Eth cutoff value was 58750000000000000000000.

src/services/merge-checker.ts
```diff
@@ -123,7 +136,18 @@ export class MergeChecker {
...
-      if (BigInt(nodeTotalDifficulty) === TERMINAL_TOTAL_DIFFICULTY && nodeBlockNumber >= MERGE_BLOCK_NUMBER) {
+      let blockchain = ''
+
+      if (ETHEREUM_BLOCKCHAIN_IDS.includes(blockchainID)) {
+        blockchain = 'ethereum'
+      } else if (GNOSIS_BLOCKCHAIN_IDS.includes(blockchainID)) {
+        blockchain = 'gnosis'
+      }
+
+      if (
```
adshmh, line 147: This needs a check of `blockchain` value: we should not look up if this key is not set

adshmh: Also, this should be covered by tests covering a the main scenarios.

adshmh (changes requested):

src/services/merge-checker.ts
crisog: > This needs a check of `blockchain` value: we should not look up if this key is not set

Addressed in 74aba5b91086c6b9ff77fc8912ae9a136ff04958

blockjoe (approved):
TDD value checks out for post merge Gnosis 👍

kutoft (approved):

adshmh (approved):

PR #1007, opened 2022-12-06
fix: reduce multi-method calls size

crisog:
T-16601


crisog:
Link T-16601

src/services/metrics-recorder.ts
```diff
@@ -116,14 +116,22 @@ export class MetricsRecorder {
         fallbackTag = ' FALLBACK'
       }
 
+      // Save raw methods for the errors db
+      const rawMethods = method
+
+      // Reduce multi-method calls for metrics/logging purposes
+      if (method && method.split(',').length > 1) {
```
kutoft: can you add a quick test for this. otherwise I think it looks good.

crisog: Added in 99d371bd1813cb10c960aa7ecdfabb76c41b7164

tests/unit/metrics-recorder.unit.ts
```diff
@@ -0,0 +1,33 @@
+import { expect } from '@loopback/testlab'
+
+describe('Metrics recorder (unit)', () => {
```
adshmh: Could you clarify the reason for these tests? They do not seem to actually test the metrics recorder code.

adshmh: Could you please add a test to cover both scenarios regarding method field? i.e. single method and multiple methods.

src/services/metrics-recorder.ts
```diff
@@ -116,14 +116,22 @@ export class MetricsRecorder {
...
+        method = 'multiple'
```
adshmh: Can we concatenate all the methods using a separator that does not increase carinality, e.g. '-' ? This could help preserve the data while also achieve the cardinality reduction.

src/services/metrics-recorder.ts
adshmh: The code will be a lot easier to read if we use the original `method` field all the way until the point it is needed in InfluxDB, at which point a new variable can be introduced and used immediately, e.g.: 

- No change until line 202, i.e. InfluxDB `new Point('relay')` call
```
simplifiedMethod = method
if (simplifiedMethod && simplifiedMethod.split(',').length > 1) {
...
```

adshmh (changes requested):
Thank you for the PR. Just a few minor blocking comments need to be addressed.

tests/unit/metrics-recorder.unit.ts
crisog: > Could you clarify the reason for these tests? They do not seem to actually test the metrics recorder code.

That means the naming needs to change. The purpose of this test was only to test the small piece reducing the methods. It could be moved as a function somewhere in the utilities if needed

src/services/metrics-recorder.ts
crisog: > Can we concatenate all the methods using a separator that does not increase carinality, e.g. '-' ? This could help preserve the data while also achieve the cardinality reduction.

How does using a different separator reduce cardinality? AFAIK Whatever separator you use, you are creating a new label as long as the method order and count is different.

Example:

2 new labels
`eth_getBlockByNumber,eth_getTransactionReceipt`
`eth_getBlockByNumber,eth_getBlockByNumber`

2 new labels too?
`eth_getBlockByNumber-eth_getTransactionReceipt`
`eth_getBlockByNumber-eth_getBlockByNumber`

src/services/metrics-recorder.ts
crisog: It was up there because the logs were supposed to be using this simplified method too. It was later decided to not too for now, so your proposed changes make sense.

Addressed in f6f49a252dc5d5b61eb1446673ddaefb226cce7d

tests/unit/metrics-recorder.unit.ts
crisog: > Could you please add a test to cover both scenarios regarding method field? i.e. single method and multiple methods.

The tests there cover both scenarios, single method (should not be reduced/simplified) and multiple method (should be reduced/simplified)

crisog: The naming solved in e924cd7db02ab031196157788add8228940920bb

adshmh (approved):
LGTM. Thanks!

PR #1009, opened 2022-12-07
fix: remove os host from influx data points

crisog:
Given how our infrastructure dynamically scales up/down, the amount of host names produced are too much for a time-series DB. This host name isn't really helpful at all on Influx, it is the most useful for logging.

This alongside #1007 should reduce cardinality by a lot.


adshmh (approved):
LGTM. Thanks!

PR #1019, opened 2023-01-09
Blockchains response endpoint

rem1niscence:
Creates a blockchains response object for the blockchains endpoint, which allows having better control of which fields are published through it

### DIFF
--- a/src/models/blockchains.model.ts
+++ b/src/models/blockchains.model.ts
@@ -183,7 +183,7 @@ export type BlockchainsResponse = {
   }[]
 }
 
-export function blockchainToBlockhainResponse(bl: Blockchains): BlockchainsResponse {
+export function blockchainToBlockchainResponse(bl: Blockchains): BlockchainsResponse {
   return {
     ticker: bl.ticker,
     hash: bl.hash,
--- a/src/models/blockchains.model.ts
+++ b/src/models/blockchains.model.ts
@@ -151,6 +151,55 @@ export class Blockchains extends Entity {
   }
 }
 
+export type BlockchainsResponse = {
+  ticker: string
+  hash: string
+  networkID: string
+  network: string
+  description?: string
+  index: number
+  blockchain: string
+  blockchainAliases: string[]
+  active: boolean
+  syncCheck?: {
+    path?: string
+    body: string
+    resultKey: string
+    allowance?: number
+  }
+  logLimitBlocks?: number
+  path?: string
+  evm?: boolean
+  redirects?: {
+    alias: string
+    domain: string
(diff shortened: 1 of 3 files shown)


src/models/blockchains.model.ts
```diff
@@ -151,6 +157,57 @@ export class Blockchains extends Entity {
...
+  logLimitBlocks?: number
+  path?: string
+  evm?: boolean
+  redirects?: {
+    alias: string
+    domain: string
+  }[]
+}
+
+export function blockchainToBlockhainResponse(bl: Blockchains): BlockchainsResponse {
```
kutoft: blockchain is misspelled in the function name

src/models/blockchains.model.ts
```diff
@@ -151,6 +157,57 @@ export class Blockchains extends Entity {
...
+    active: bl.active,
+    syncCheckOptions: {
+      path: bl?.syncCheckOptions?.path,
+      body: bl?.syncCheckOptions?.body,
+      resultKey: bl?.syncCheckOptions?.resultKey,
+      allowance: bl?.syncCheckOptions?.allowance,
+    },
+    enforceResult: bl?.enforceResult,
+    logLimitBlocks: bl?.logLimitBlocks,
+    path: bl?.path,
```
kutoft, line 205: why is bl possibly null/undefined here but not in the first few with ticker and hash, etc...

rem1niscence: because these values are not on all the chains so they might be null, for example path is only used for AVAX and log limit blocks is not on all chains either

kutoft (dismissed):

tests/acceptance/blockchain.controller.acceptance.tests.ts
```diff
@@ -80,10 +76,39 @@ describe('Blockchains controller (acceptance)', () => {
 
   it('retrieves the information of a specific blockchain', async () => {
```
adshmh, line 77: Could we add test cases for other scenarios? at-least the following should be covered IMO: incorrect blockchain hash, any field from the blockchain object that can be null/empty.

adshmh (changes requested):
Thank you for the PR! Just one blocker to address.

tests/acceptance/blockchain.controller.acceptance.tests.ts
rem1niscence: added test for incorrect hash. As for field or null empty fields are already handled on the other tests implicitly, not all the fields are used

adshmh (approved):
LGTM. Thanks!

PR #1036, opened 2023-02-01
T-18303 Use phd v1

commoddity:
This PR simply adds the `v1` path to the PHD client.

Integration tests confirm it is getting data as it should from PHD.


src/services/phd-client.ts
```diff
@@ -125,7 +123,6 @@ class PHDClient {
       }[path]
 
       const processedDocument = processMethod?.() || document
-      console.log('ERROR', { processedDocument, modelFields })
```
adshmh, line 128: Why does this need to be removed?

adshmh: NVM. Seems like it shouldn't have been here in the first place.

adshmh (dismissed):
LGTM. Thanks!

rem1niscence (approved):

PR #1037, opened 2023-02-06
Remove mongodb

rem1niscence:
Remove all references of mongodb to only keep the postgres db as the only source of truth. This is done by
* Removing anywhere where mongo was used
* Removing fallback code from phd
* Modify all tests that needed to have some data from a db to use a HTTP mock to retrieve PHD results
* Add a small function to phd to return HTTP error objects in case the HTTP call fails

Although I tried my best to not modify any logic, a small change had to be done to the `loadBlockchains` function to return the expected `Error` response, besides that no other change was made to any logic


tests/unit/pocket-relayer.unit.ts
```diff
@@ -275,12 +266,13 @@ describe('Pocket relayer service (unit)', () => {
   })
 
   it('throws an error when loading an invalid blockchain', async () => {
+    axiosMock.reset()
```
adshmh, line 269: Why is this needed here? Doesn't this happen after each test case already with the added code?

commoddity (approved):
These changes all look good to me. Thanks man.

adshmh (approved):

tests/unit/pocket-relayer.unit.ts
rem1niscence: yes and that's exactly why I put it there, to remove the blockchain response so it returns 404 and mimics the not found test

PR #1043, opened 2023-02-23
T-18925 fix: removed unneeded checks for required fields from PHD Client

commoddity:
Removes the check for required model fields in the PHD client as it's no longer relevant since the move to remove mongo (and causes hard to maintain issues if/when we change DB fields).


rem1niscence (dismissed):
LGTM! always like PRs that remove stuff

commoddity:
> LGTM! always like PRs that remove stuff

Same. The best kind of PRs.

rem1niscence (approved):

adshmh (approved):

